PR #6, opened 2015-02-16
fix per-VM alsa config path

jean-edouard:
OXT-191

Singed-off-by: Jed Lejosne <jed.openxt@gmail.com>

### DIFF
--- a/xenmgr/Vm/React.hs
+++ b/xenmgr/Vm/React.hs
@@ -277,7 +277,7 @@ whenShutdown xm reason = do
     return ()
     where
       removeAlsa domid = liftIO $ do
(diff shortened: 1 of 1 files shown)


rossphilipson:
I know this is part of a larger PR set but this looks like a real fix on its own. Can this one just be pulled as is?

jean-edouard:
Yes, the per-VM alsa config files now live in /var/run.
Every place that uses /tmp is wrong, including this one.
This indeed is tied to other pull requests in xenclient-oe.git and toolstack.git, but pulling it on its own shouldn't hurt.
Thanks!

PR #8, opened 2015-03-30
Rebase PV Linux driver on standard PVOPS kernel/xenbus.

rossphilipson:
Get rid of the CPUID that allowed the xc-xen platform
device to load.

OXT-231


eric-ch:
LGTM, +1 (with https://github.com/OpenXT/pv-linux-drivers/pull/1).

PR #12, opened 2015-05-07
If VM has multifunction gpu passthrough, pass all functions on device.

crogers1:
For AMD cards to function properly, they must support bus level reset.
Bus level reset requires all functions to be passed to the guest, so
at VM boot time, check for GPU passthrough and ensure extra functions
are also passed through.

OXT-276

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -191,6 +191,7 @@ import XenMgr.CdLock
 import {-# SOURCE #-} XenMgr.PowerManagement
 import Rpc.Autogen.XenmgrNotify
 import XenMgr.Expose.ObjectPaths
+import Vm.Pci
 
 data EventHook
    = EventScript FilePath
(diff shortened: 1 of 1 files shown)


jean-edouard:
If the aim is to do a bus reset, shouldn't we pass all the devices on the bus (and their functions) to the guest, instead of just all the functions from the GPU card?

jean-edouard:
Also I'm pretty bad at Haskell, but I'm sure isGpuPt can have a more elegant return statement :)

jean-edouard:
Something like that maybe?
isGpuPt uuid = let gpu = getVmGpu uuid in gpu != "" && gpu != "hdx"

crogers1:
I'm not sure in what case we could have more than 1 device on the same bus that the GPU is using.

getVmGpu returns a string in a monadic context, so I need it bound to that gpu name so I can work with the string itself.  I need the returns to lift the bool back into the outer monadic context.  The 'do' and if-else syntax I used might be a little verbose, but I think it's more readable that way.

rossphilipson:
So it may be unlikely that a secondary bus has more than one device on it but it is certainly possible. In that case, all devices and functions must be co-assigned to the guest. The other option is to handle that as an unsupported configuration.

crogers1:
Ah, ok.  No I don't think there's a need to make it unsupported.  If we want to pass through all the devices, I suppose I just need to strip off the device and the function in stripBdf so it matches on "Domain:Bus:*.*" instead of just "domain:bus:device.*"

rossphilipson:
love the bus

crogers1:
Updated, implemented Ross and Jed's suggestions.

jean-edouard:
stripPositional looks a lot like "take" to me :)

crogers1:
Yep.  Went from a more specific solution to something more generic and viola, we have a Prelude function...

jean-edouard:
The code looks better now.

However, should we do that for all GPUs?
GPUs that have FLR reset capabilities don't need their whole bus with them to be reset-able.

Maybe we just want to passthrough the whole bus all the time anyway for consistency?
Or maybe that could be a system advanced option?

crogers1:
Passing through the whole bus for all GPUs does seem heavy-handed.  In my experience, heavy-handedness is usually not good. 

However, I feel that the case where a user wants to pass the video component to one guest and the audio to a separate guest is rare.  We might even want to declare that as an unsupported configuration since this is something we're trying to avoid, especially for AMD cards.  I could be wrong though, in which case perhaps rather than passing the rest of the bus through at VM start time, maybe do it when the GPU is assigned for passthrough and let the user deal with unassigning the extra devices/functions.  Without enforcing the passthrough on start though, we run the risk of the user doing something stupid.

Regardless, we need to decide on a policy soon.

transpute:
Whatever we decide, the UI and/or API should report to the user/caller what is being passed through, i.e. side effects of the command.

pearsonk:
I agree with @transpute that the toolstack/UI should report any/all devices that are consequently passed when a user passes a particular device (if any). It should be clear to a user what effects the action has on the system.

I personally don't see a use case for splitting video/audio among different VMs, but quite often I pass the video portion and leave the audio unassigned so I don't have so many devices to wade through in the guest. Since this is limited to GPU for the moment, I don't see a problem but I can think of at least 1 NIC where this would have caused problems. Where's SR-IOV for everything when you need it...

rianquinn:
Whether there is a use case for assigning split devices to more than one VM simultaneously irrelevant. If the toolstack has to do a bus level reset, you cannot do PCI pass-through without passing all of the devices to the VM. Any attempt to do so will crash the VM/hypervisor once the reset occurs. 

Also, putting the logic in the toolstack during device setup (i.e. writing the settings to the db) will break OTAs. It also prevents the user from assigning split devices to VM that will not be used simultaneously. I agree though, there should be a UI modification to tell the user that the following devices will also be passed through during VM boot, just like there should be a UI mod that tells the user when a device is assigned to more than one VM  (which should now include devices that as pass-through due to this issue).

rossphilipson:
Actually you will only crash the system if the other device is not passed through and a driver is loaded for it in dom0. That is why Kevin would be able to leave the sound cards behind. That said, it really is best practice to co-assign in these cases.

"Where's SR-IOV for everything when you need it.." <-- or just flippin' FLR - how hard can it be...

rianquinn:
I could not agree more with FLR. It's amazing to me how many devices don't have this. I guess the main reason that cheap nVidia cards don't work is because they refused to support FLR (or any form of reset). 

Yeah, at the moment, I don't know of any logic the specifically prevents the drivers from being loaded (there might be). I think that would be pretty complicated as there would have to be some way of allowing some drivers to be loaded for devices that are shared, and then preventing drivers from being loaded for devices that should be passed through.

rossphilipson:
Yea it is simply that the driver doesn't exist e.g. with the ATI cards. But there is nothing specific and if the other devices.fns are driven, they will explode. So it is best practice.

jean-edouard:
LGTM, will merge soon

PR #13, opened 2015-05-28
Implement proper monadic evaluation and filtering with mapM_

crogers1:
Vm/Actions.hs

  * Use mapM_ to force evaluation of all Rpc monadic computations
    to make sure we add every function/device that matches.
  * Additionally ensure we filter out the video function of gpu to
    avoid passing it through twice.

OXT-296


jean-edouard:
+1, tested, works fine.

PR #15, opened 2015-07-16
Allow Stubdomain configuration of a given VM through the toolstack

eric-ch:
Add logic to xenmgr for:
 - stubdom-cmdline:
   * cmdline to be appended to the stubdomain's default cmdline.
     Can be set from the VM configuration file or through DBus
 - stubdom-memory:
   * Memory given to the stubdomain, in megabytes.
     Can be set from the VM configuration file or through DBus

OXT-332

See related PR in:
- xenclient-oe.git
- toolstack.git
- idl.git


crogers1:
+1, looks good to me!

jean-edouard:
+1

eric-ch:
Defines the "default" stub-domain kernel command line in the VM configuration templates. Also add stubdom-memory argument to make it more visible.

cjp256:
@eric-ch I don't see it, but perhaps I am missing something. Is there a default specified in the toolstack if unspecified?  I'm concerned about VMs that will likely break after an upgrade.

eric-ch:
@cjp256 No I removed the default in https://github.com/eric-ch/toolstack/commit/bd0d4579282bedc15711f1ed2fd535d01a31c478, which is not good for existing VMs at the time of the upgrade. We might as well keep the default and let the configuration override it.

cjp256:
@eric-ch : sounds like a good idea.

eric-ch:
@cjp256 https://github.com/eric-ch/toolstack/commit/57d13ae452d4d78185a25b746544cbbf1462d188 default value is back.

PR #17, opened 2015-08-12
Added rpc-firewall-rule for permitting xcpmd

Marsland:
Edited service-uivm template to permit xcpmd messages to support
conversion from UPower to xcpmd in the UI


jean-edouard:
Nice find, I've been looking for that rule!
+1, will merge soon

Marsland:
Chris and I tracked the problem I had with the code not working only on the actual laptop to missing the firewall rule here. Doc will be making a subsequent change to the same file to support DispMon DBus messages as well.

PR #19, opened 2016-03-29
[xenmgr] Remove the unused dbd2 Haskell implementation.

rossphilipson:
I agree that removing it is a good idea. It is confusing and we
certainly don't want more Haskell code ever getting into OpenXT.

OXT-467


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/30

jean-edouard:
Build and installed without trouble, merging soon.

PR #23, opened 2016-05-25
Fix shutdown/reboot behavior from sleep.

crogers1:
Vm/Actions.hs:
  -shutdownVm: Add threadDelay of 5 seconds to give domain time to resume
   if the domain was previously in S3 before issuing the shutdown command.

  -rebootVm: Add threadDelay of 5 seconds for reboot case as well

OXT-354

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -916,12 +916,18 @@ disconnectFrontVifs back_uuid =
 rebootVm :: Uuid -> Rpc ()
 rebootVm uuid = do
     info $ "rebooting VM " ++ show uuid
+    acpi <- getVmAcpiState uuid
     -- Write XENVM configuration file
     writeXenvmConfig =<< getVmConfig uuid True
 
     -- Ask xenvm kindly to reload it
     Xenvm.readConfig uuid
 
+    when (acpi == 3) $ do
+      info $ "resuming " ++ show uuid ++ " from S3 first.."
+      resumeS3AndWaitS0 uuid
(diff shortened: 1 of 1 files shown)


pearsonk:
I'm usually wary about arbitrary sleep-waits to address a race condition. Is there any deliberate reasoning behind 5 seconds? Or is just a simple approach to reduce the problem, but not necessarily eliminate the possibility of recurrence?

crogers1:
5 seconds seemed sufficient to ensure the guest OS was fully resumed from S3.  From the toolstack perspective, I can't find anything else to key off of.  ACPI state (which it is waiting on currently) reports 0 before the guest is fully resumed, thereby ignoring the shutdown request.  This patch is the simplest solution (and smallest footprint) I could think of.

An alternative (but potentially larger change) is to close this PR and fix the problem in the PV driver, retrying the shutdown until a timeout, or perhaps asking Windows if it ok to shutdown, though I'd have to see if that function exists.  Either way, I think the solution is going to be a wait/retry, as the problem is a timing issue.

The waitResumeS3 for the reboot case still needs to be added though, as the current behavior of sleep -> reboot just hangs forever.

richpersaud:
*If* we are going to use a static delay, we should test this on a machine where there is another guest that is actively using most of the CPU, to estimate an upper bound on the time needed to resume the guest that is about to be shutdown.

crogers1:
I'll spend a little more time this morning before the triage call to see if I can't come up with something better.  I agree that sleeps are ugly.

crogers1:
Closing for now, will reopen with different approach.

PR #24, opened 2016-05-26
templates: avoid wildcard firewall rules to dom0

jean-edouard:
OXT-558


eric-ch:
LGTM

PR #27, opened 2016-06-02
Fix shutdown/reboot behavior from sleep.

crogers1:
Vm/Actions.hs:
  -Add a function that checks to see if the PV driver xs watches
   are responding after resuming from S3. Shutdown and reboot
   commands attempt to see if the drivers are listening before writing
   the "control/shutdown" node.

OXT-354

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -911,19 +911,49 @@ disconnectFrontVifs back_uuid =
           info $ "disconnecting nic uuid=" ++ show front_uuid ++ " id=" ++ show nic_id
           Xenvm.connectVif front_uuid nid False
 
+xsWaitForNodeToDisappear :: Int -> String -> IO Bool
+xsWaitForNodeToDisappear timeout_secs node = do
+    r <- timeout (10^6 * timeout_secs) $ xsWaitFor node test
+    case r of
+      Just () -> return True
+      Nothing -> return False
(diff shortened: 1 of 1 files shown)


crogers1:
Linked to https://github.com/OpenXT/xc-windows/pull/29

This is better than the arbitrary sleep I had proposed before.  I really wish I had the [RegisterResumeSuspendNotification](https://msdn.microsoft.com/en-us/library/windows/desktop/jj613321%28v=vs.85%29.aspx) function, but it's only available in Windows 8.  Instead, wait until the pv drivers respond by deleting the "control/awake" node, signalling they are in a state where they can fire off a xenstore-watch callback, and therefore able to issue a shutdown command when the "control/shutdown" node is written when calling shutdownVm or rebootVm.

jean-edouard:
Isn't this sleeping 10 times 1 seconds instead of 2 times 5 seconds in https://github.com/OpenXT/manager/pull/23? :)

crogers1:
#23 was a dumb sleep.  This is a timeout.  There are timeouts all over the toolstack, both in xenmgr and xenvm.  I'm trying to create an event to key off of (ie, waiting for the control/awake node to disappear, before sending the shutdown command) since acpi state change event doesn't work.

I haven't had time to look at Ross' code yet, which might provide a better event to key off of.

rossphilipson:
Closing for now.

PR #28, opened 2016-06-11
[v4v] Remove domU access ports to dom0

rossphilipson:
Now that the win-tools are gone that used these ports, they can be closed
down.

OXT-563

### DIFF
--- a/templates/default/new-vm
+++ b/templates/default/new-vm
@@ -1,12 +1,10 @@
 {
   "type": "svm",
   "v4v-firewall-rules": {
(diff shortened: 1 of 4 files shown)


rossphilipson:
Started a build with this change:
http://158.69.127.133:8020/builders/openxt-custom/builds/89

jean-edouard:
LGTM, build and tested.
Noticed guest warnings with Windows 10, but also happens without this PR. (see OXT-606)
Merging soon.

PR #29, opened 2016-06-15
Fix templates broken by kernel filename change and fix OTA upgrade cases by adding associated db migrations

cjp256:


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/92

jean-edouard:
LGTM, merging soon.

PR #30, opened 2016-06-16
Replace hvm-info with xenops physinfo instead.

eric-ch:
xenops (CAML toosltack) provides access to xc_physinfo without the need
for a standalone binary, so use that instead of hvm-info.

OXT-586


jean-edouard:
This is still calling an external binary though.
What is the motivation here? Getting rid of `hvm-info`?

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/92

jean-edouard:
Works fine, hvm-info is a patch we have against Xen that we'll be able to drop once this is merged.
Merging soon.

PR #32, opened 2016-06-27
Add migration script to disable cdrom autolock

jean-edouard:
OXT-635

### DIFF
--- a/upgrade-db/Migrations.hs
+++ b/upgrade-db/Migrations.hs
@@ -49,6 +49,7 @@ import qualified Migrations.M_25
 import qualified Migrations.M_26
 import qualified Migrations.M_27
 import qualified Migrations.M_28
+import qualified Migrations.M_29
 
 migrations :: [Migration]
 migrations = [ Migrations.M_1.migration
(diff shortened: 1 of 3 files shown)


rossphilipson:
Things seem to work fine with non-USB CD devices. No auto assignments happens but explicit assign and un-assign work. USB CD devices do still get auto assigned to the VM in focus. Not sure what the intended behavior should be...

jean-edouard:
Yes, that's expected, I'll make a PR to prevent auto-assignment of USB optical devices to make things more obvious.

rossphilipson:
The device UI seems to be broken or at least got broken when it tried to deal with a USB CD device. It is stuck in a mode where it cannot assign devices. I have no idea what is wrong - I will make a separate ticket for that.

Given the problems above and that this was originally about CD assignment and not USB assignment (which works fine), I would argue for leaving USB assignment alone for the moment. The issue this PR addresses was working before I started fooling with USB devices so I am going to merge the original PR soon.

rossphilipson:
I think that changing USB assignment behaviors is not really part of this ticket and should have its own (and possibly an RFC on the ML).

PR #33, opened 2016-08-22
Move the USB mouse rule to a new migration script

jean-edouard:
OXT-582


dozylynx:
OXT-729 filed to track the upgrade fix, as separate from the original OXT-582 issue.

rossphilipson:
These both LGTM - merging soon

PR #35, opened 2016-08-23
Handle the default stubdom memory and cmdline outside of the VM configs

jean-edouard:
OXT-730

### DIFF
--- a/templates/default/new-vm
+++ b/templates/default/new-vm
@@ -25,8 +25,6 @@
     "display": "none",
     "boot": "cd",
     "flask-label": "system_u:system_r:hvm_guest_t",
-    "stubdom-memory": "80",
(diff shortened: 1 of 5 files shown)


eric-ch:
LGTM, +1.

eric-ch:
Will merge soon.

rossphilipson:
@eric-ch are you ok with losing the ability to change the stubdom command line (or maybe I am missing something)?

@jean-edouard should this go in stable 6 too?

eric-ch:
@rossphilipson You can still change it using the configuration file. It just now uses the default value hardcoded in xenmgr & xenvm by default. Since this is a value aimed mostly for testing new stubdoms, it seems fine to me.

rossphilipson:
@eric-ch ack that smack

PR #36, opened 2016-08-25
STABLE-5: Handle the default stubdom memory and cmdline outside of the VM configs

jean-edouard:


rossphilipson:
OXT-730

PR #41, opened 2016-09-08
[ota] STABLE-6 Don't download tarballs twice.

rossphilipson:
Use the forground step to download the meta files and the background
step to do the main download. This also fixes some oddities with the UI.

OXT-734


jean-edouard:
Worked perfectly, merging soon.

PR #43, opened 2016-11-21
libXL support in Xenmgr

crogers1:
This series of commits provides support for libXL as the lower level toolstack piece, replacing xenvm.  The PR has been broken up into 4 commits and the changes are documented in each commit message.


jean-edouard:
Fails to build, see http://158.69.127.133:8020/builders/openxt-custom/builds/215

jean-edouard:
The previous comment was a bug in the idl that's now fixed.

templates/default/new-vm
```diff
@@ -4,7 +4,8 @@
     "0": "my-stubdom -> 0:4001",
     "1": "my-stubdom -> 0:4002",
     "2": "my-stubdom -> 0:5000",
-    "3": "my-stubdom -> 0:5001"
+    "3": "my-stubdom -> 0:5001",
+    "4": "my-stubdom -> 0:5100"
```
jean-edouard: Unfortunately this change will need a migration script to handle upgrades...

templates/default/service-uivm
```diff
@@ -66,7 +66,7 @@
     "hvm": "false",
     "apic": "true",
     "nx": "true",
-    "vkbd" : "true",
+    "vkb" : "true",
```
jean-edouard: Same as above, if we really want to rename this attribute we need a migration script.
Maybe it's best to just keep the old name?

xenmgr/Vm/Actions.hs
```diff
@@ -794,31 +828,65 @@ bootVm config
           liftIO $
             mapM_ updateCdDeviceMediaStatusKey =<< liftIO getHostBSGDevices
 
+      setupBiosStrings uuid =
+          whenDomainID_ uuid $ \domid -> do
+            liftIO $ xsWrite (xsp domid ++ "/bios-strings/xenvendor-manufacturer") "OpenXT"
+            liftIO $ xsWrite (xsp domid ++ "/bios-strings/xenvendor-product") "OpenXT 5.0.0"
```
jean-edouard: Dunno what that is but the version number seems off...

crogers1:
Addressed the above comments with the latest commit on this branch

crogers1:
Fixed a typo in service-uivm which causes the platform to lose mouse and keyboard input.

templates/default/new-vm
rossphilipson: So will the next 3 also it seems. Let's put a ticket in so that doesn't get lost.

rossphilipson:
Merging soon.

PR #44, opened 2016-11-22
[Upgrade] Allow wildcard matching on upgradeable versions.

rossphilipson:
Thanks to Chris Rogers for the code.

OXT-832

Signed-off-by: Ross Philipson <philipsonr@ainfosec.com>

PR #45, opened 2016-11-22
[Upgrade] STABLE-6 Allow wildcard matching on upgradeable versions.

rossphilipson:
Thanks to Chris Rogers for the code.

OXT-832


rossphilipson:
Tested both the success case with an original 6.0.1 and a case that should fail where the original version was 6.1.0. Both behaved correctly.

jean-edouard:
See https://github.com/OpenXT/manager/pull/44#issuecomment-263318931

PR #48, opened 2016-12-15
xenmgr: handle empty power management actions as "nothing"

jean-edouard:
and log wrong values

OXT-866

### DIFF
--- a/xenmgr/XenMgr/PowerManagement.hs
+++ b/xenmgr/XenMgr/PowerManagement.hs
@@ -111,7 +111,8 @@ pmActionOfStr "shutdown"  = ActionShutdown
 pmActionOfStr "forced-shutdown" = ActionForcedShutdown
 pmActionOfStr "reboot"    = ActionReboot
 pmActionOfStr "nothing"   = ActionNothing
-pmActionOfStr _           = error "incorrect pm action specification"
+pmActionOfStr ""          = ActionNothing
+pmActionOfStr bad         = error ("incorrect pm action specification: " ++ bad)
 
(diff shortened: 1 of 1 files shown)


pearsonk:
This fixes the "" handling causing problems, but would this cause similar problems on bad or erroneous PM values with the error() call? I think it should be logged, but configuration errors in the PM policy shouldn't have the power to cause such crippling in the system. Ideal behavior would be that it logs an unknown/inappropriate value but takes the "do nothing" action.

I'm haven't looked deeply in the code, so ignore this post if my assumptions are wrong.

jean-edouard:
@pearsonk this is a good idea, but I'd need some help to figure out how to log an error *and* keep going with "nothing"...

crogers1:
Unfortunately, this function isn't being used in the IO context, so the only way to perform IO (ie logging the error) on demand like that is to use _unsafePerformIO_ which isn't recommended, hence the function name.

I woudn't use _error_ at all, since it's supposed to stop execution.  Just have invalid PM action strings evaluate to ActionNothing.

jean-edouard:
Silently ignoring bad values doesn't sound great to me...
It could hide issues, like if the UI gets rewritten with a typo in a pm state or something.
Haskell is rea^W^W^W

crogers1:
With some finesse you could probably get it to log an error, but it would be a larger diff.  If you want I can spend a little time tomorrow looking at it.

jean-edouard:
Replaced by https://github.com/OpenXT/manager/pull/49

PR #50, opened 2016-12-19
updatemgr: checkVersion: write end conditions

jean-edouard:
OXT-874


crogers1:
+1. Empty list is fine to use as terminating condition, since the recursive call only happens on x == y anyway.

rossphilipson:
Merging soon to test

rossphilipson:
Tested stable 6 version with custom master set to upgradable from 6.0.1. Worked fine, merging soon.

PR #53, opened 2017-01-30
[libxl] Fix bad config option for disabling usb

crogers1:
OXT-435


crogers1:
Xl style syntax requires 0 or 1 for booleans.  "usb=false" fails to parse config, producing IO error and failure for domain to start.

jean-edouard:
Yup, confirmed here: https://xenbits.xen.org/docs/4.6-testing/man/xl.cfg.5.html#SYNTAX
Merging soon.

PR #55, opened 2017-01-31
Replace a xenops call by an xl one

jean-edouard:
OXT-916


rossphilipson:
Build here (testing now):
http://openxt.ainfosec.com/builds/custom/master/584/

PR #57, opened 2017-02-06
[libxl] Wrap xl failures into special Exception

crogers1:
When our xl calls return failure, we throw from our new Exception
  class.  Due to the monadic context these calls are executed in,
  they are caught and wrapped in the generic "IO Error" defined in
  Errors.hs. The new Exception class allows us to pass along a
  string description of the error so that it can be displayed
  in the UI or logs, depending on where the call originated. This
  should fix the cryptic, non-descriptive "IO Error." message when
  domain operations such as start fail.

  OXT-435

### DIFF
--- a/xenmgr/XenMgr/Connect/Xl.hs
+++ b/xenmgr/XenMgr/Connect/Xl.hs
@@ -10,7 +10,7 @@
 --
 
 
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE OverloadedStrings, DeriveDataTypeable #-}
 module XenMgr.Connect.Xl
     (
     --xl domain control
@@ -56,6 +56,7 @@ import Control.Monad
 import Control.Monad.Error hiding (liftIO)
 import Control.Concurrent
 import Data.String
+import Data.Typeable
 import Data.Text as T
 import Vm.Types
 import Vm.DmTypes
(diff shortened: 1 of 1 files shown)


crogers1:
This is linked to https://github.com/OpenXT/toolstack-data/pull/42

xenmgr/XenMgr/Connect/Xl.hs
```diff
@@ -69,17 +70,25 @@ import System.Process
...
 
 type NotifyHandler = [String] -> Rpc ()
 type Params = [(String, String)]
 
+data XlExceptionClass = OOMException String | XlException String
+    deriving (Typeable)
+instance Exception XlExceptionClass
+
+instance Show XlExceptionClass where
+    show e@(XlException s) = show s
```
jean-edouard: This function definition is non-exhaustive and doesn't compile:

    | XenMgr/Connect/Xl.hs:84:4:
    |     Warning: Pattern match(es) are non-exhaustive
    |              In the definition of `show': Patterns not matched: OOMException _

crogers1: Ah. I'll fix that up tomorrow.

crogers1:
Compiler error fixed.  Removed one of the entries in my instance for Show for some reason.

rossphilipson:
Starting another build of all this.

rossphilipson:
While testing we saw an out of memory exception thrown for something that was clearly not an OOM situation. It seems the logic for this is not quite right yet.

crogers1:
It seems that xl return codes are not even close to being unique.  I probably need to go through xl_cmdimpl.c in the libxl tree and put in more descriptive return values.

crogers1:
Going to close these PRs until I can come up with a better solution.

PR #58, opened 2017-02-07
[libxl] Fix bug with cdrom access

crogers1:
If a domain disables cdrom access, the "-drive" entry is still
  placed in the device_model_params, which, by itself, fails to
  parse in qemu causing the domain to fail to boot.

  OXT-435


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/588
LGTM, merging soon.

PR #59, opened 2017-02-23
xenmgr: match all GPUs (class == 0x03..)

jean-edouard:
OXT-962

### DIFF
--- a/xenmgr/Vm/Pci.hs
+++ b/xenmgr/Vm/Pci.hs
@@ -52,6 +52,7 @@ import Data.Char
 import Data.Function
 import Data.List
 import Data.Maybe
+import Data.Bits
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Char8 as BC
 import qualified Data.Map as M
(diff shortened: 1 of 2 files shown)


jean-edouard:
GPUs don't necessarily have a device class of 0x0300 (we just found one that's 0x0380).
Anything that starts with 0x03 is a GPU.
See http://lxr.free-electrons.com/source/include/linux/pci_ids.h?v=4.9#L35

MagOXT:
Tested on HP EliteDesk 800 G1. Drop down available for passthrough. 
PR OK

rossphilipson:
Yea the second byte is the subclass (and usually pretty meaningless). E.g. that subclass is listed as "	Other display controller" in a couple of places. Unfortunately I cannot understand any of the code change...

crogers1:
Instead of duplicating the function to add the predMask line, we should change gpuPciPtRule to mimic the amtPciPtRules (ie make it a list).  The lxr link only has 4 class id's defined, and I can't imagine that header gets updated often.  So for now, it seems acceptable to enumerate them all since 0x380 is a catchall class 'PCI_CLASS_DISPLAY_OTHER'.  That means this function gets changed from
```haskell 
pciGetGpus = mapM (pciGetDevice . fst) =<< matchingPciAddresses gpuPciPtRule
```
to
```haskell
pciGetGpus = mapM (pciGetDevice . fst) =<< matchingPciAddressesMany gpuPciPtRules
```
and gpuPciPtRule gets changed to
```haskell
 gpuPciPtRules :: PciPtRule
 gpuPciPtRules = 
     [ PciPtRule (Just 0x300) Nothing Nothing True
     , PciPtRule (Just 0x301) Nothing Nothing True
     , PciPtRule (Just 0x302) Nothing Nothing True
     , PciPtRule (Just 0x380) Nothing Nothing True
     ]
```
Just my opinion, this way we follow the existing implementation for multiple pciRules and avoid code duplication.

pearsonk:
I believe that @crogers1 approach is cleaner, but has a common whitelisting deficiency that it could cause anything not explicitly listed to not match, even though it is of class 0x03XX. pci-ids is neither comprehensive, nor exhaustive, so the simpler solution could cause failure to match devices that otherwise could pass through just fine. I can't qualify the likelihood of this occuring, though, but I would assume it would be pretty rare.

jean-edouard:
Yup, my fix is ugly, but it does the right thing.
Limiting the match to 4 values is virtually just as bad as the previous 1 value match IMO.

crogers1:
How about doing it @jean-edouard 's way, but instead of copying the whole function, add predMask() to the original function and do something with a logical OR such as
```haskell
predAll entry = (pred ruleClass pciinfoClass entry  || predMask ruleClass pciinfoClass entry )&& ...
```
So that if pred() fails because the class isn't defined in our rules (ie. something other than 0x300), predMask() catches the case and returns true.  So we have a limited code change, we fix the broken case and still support the existing code that calls matchingPciAddresses elsewhere in xenmgr.

jean-edouard:
@crogers1 won't that also mask other devices? (like match wireless nics (0x0280) when only looking for wired NICs (0x0200))

crogers1:
Yea it would.  How about implementing an 'if' or 'case' block that masks out the digit we care about and compare it against 3, since the original commit only applies it to gpu devices anyway?

jean-edouard:
That's pretty much what I tried to do... You're welcome to PR something more elegant, I'll review it!

jean-edouard:
Replaced by https://github.com/OpenXT/manager/pull/62

PR #61, opened 2017-02-23
[xenmgr] Fix naming scheme for temp vhds

crogers1:
When disk persistence is set to false, we use non-encrypted
  snapshots.  If the parent disk is encrypted and the snapshot
  shares the same name, we can fall into an error case in blktap2
  crypto code meant to catch when an encrypted disk is replaced
  with a blank disk. This commit prepends to the snapshot name so
  it's unlikely we hit this case.

  OXT-966


crogers1:
replaced by #70

PR #62, opened 2017-02-24
[xenmgr] Gpu pci class id masking

crogers1:
OXT-962

### DIFF
--- a/xenmgr/Vm/Pci.hs
+++ b/xenmgr/Vm/Pci.hs
@@ -49,6 +49,7 @@ module Vm.Pci (
            ) where
 
 import Data.Char
+import Data.Bits
 import Data.Function
 import Data.List
 import Data.Maybe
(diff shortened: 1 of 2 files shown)


crogers1:
After discussing alternate approaches with @jean-edouard, this is the one we settled on.  By only calling predMask when looking for gpu devices, we match on all devices of 0x3XX, thereby solving the original problem without code duplication.  We maintain the original behavior for other device types, so distinctions between device classes such as wired and wireless nics are observed.

pearsonk:
I like the approach and reusability of this PR. Easy to understand, for Haskell.

jean-edouard:
+1, closing mines

rossphilipson:
Merging soon since it is identical to https://github.com/OpenXT/manager/pull/63 and that was tested.

PR #67, opened 2017-03-07
[xenmgr] Simplify acpi table code

crogers1:
OpenXT only needs to support passing one acpi table to the guest.
  Therefore, rename the property to something more sensible, return
  it back to a bool value, and include the path to the table in the
  xl config generation code.

  OXT-748


rossphilipson:
This is working. SLIC is passed in when allow OEM installs is set. Merging soon with other 2.

PR #68, opened 2017-03-08
[libxl] Use special exception class and capture

crogers1:
stderr from libxl. For domain creation, we now capture stderr from
  libxl and return it directly to the caller, as it might fail for
  a variety of reasons. Other toolstack calls that fail are more
  predictable and can simply provide a short error message to
  the caller.

  OXT-991

### DIFF
--- a/xenmgr/XenMgr/Connect/Xl.hs
+++ b/xenmgr/XenMgr/Connect/Xl.hs
@@ -10,7 +10,7 @@
 --
 
 
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE OverloadedStrings, DeriveDataTypeable #-}
 module XenMgr.Connect.Xl
     (
     --xl domain control
@@ -56,6 +56,8 @@ import Control.Monad
 import Control.Monad.Error hiding (liftIO)
 import Control.Concurrent
 import Data.String
+import Data.List as L
+import Data.Typeable
 import Data.Text as T
 import Vm.Types
 import Vm.DmTypes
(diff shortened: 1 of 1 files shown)


xenmgr/XenMgr/Connect/Xl.hs
```diff
@@ -188,10 +199,14 @@ start uuid =
...
         Shutdown -> do
-                      exitCode <- system ("xl create " ++ configPath uuid ++ " -p")
-                      case exitCode of
+                      (_, _, Just err, handle) <- createProcess (proc "xl" ["create", configPath uuid, "-p"]){std_err = CreatePipe}
+                      ec <- waitForProcess handle
+                      stderr <- hGetContents err
+                      case ec of
                         ExitSuccess -> return ()
-                        _           -> error "error creating domain"
+                        ExitFailure rc ->
```
jean-edouard: Why not just that?

    ExitFailure _ -> throw $ XlException $ L.intercalate "<br>" $ L.lines stderr

crogers1: I'll be honest I don't remember writing that.  Thanks for the heads up, that ExitFailure case can be just _ -> throw ... instead

jean-edouard:
Code LGTM, will add to custom build.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/689

jean-edouard:
Looks good, thanks, merging soon.

PR #74, opened 2017-03-20
Vwif

crogers1:
Implement vwif support in xenmgr.

OXT-973

### DIFF
--- a/xenmgr/Vm/Monitor.hs
+++ b/xenmgr/Vm/Monitor.hs
@@ -46,11 +46,14 @@ import Tools.Misc
 import Tools.Text
 
 import Vm.Types
+import Vm.Config
 import Vm.State (stateFromStr)
 import Vm.Queries
 import Vm.ConfigWriter
+import Vm.Utility
 import qualified XenMgr.Connect.Xl as Xl
 import qualified XenMgr.Connect.GuestRpcAgent as RpcAgent
+import qualified XenMgr.Connect.NetworkDaemon as ND
 import XenMgr.Rpc
 import XenMgr.Errors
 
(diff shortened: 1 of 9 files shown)


jean-edouard:
Didn't work for me... After starting a VM and connecting the laptop to the wifi, the disconnect node is still at 1.

xenmgr/Vm/Monitor.hs
```diff
@@ -134,15 +137,19 @@ submitVmEvent m e = liftIO $ (vmm_submit m) e
 
 insertDefaultEvents :: VmMonitor -> Rpc ()
 insertDefaultEvents m = let uuid = vmm_uuid m in do
+    isNdvm <- readConfigPropertyDef uuid vmType ""
```
jean-edouard: Looks like this value is more `vmType` than `isNdvm`, no?

xenmgr/Vm/Monitor.hs
```diff
@@ -160,22 +167,33 @@ insertDefaultEvents m = let uuid = vmm_uuid m in do
           _ -> return ()
     whenPowerState _ = return ()
 
+    whenNetState state backend = 
+        case state of
+            30  -> manageFrontVifs False (ND.netbackToUuid backend) --Network backend disconnected, handle it...
+            100 -> manageFrontVifs True (ND.netbackToUuid backend)
```
jean-edouard: What are these numbers? Are they documented somewhere?

crogers1: They're integers network manager uses to enumerate device state.  The UI was already hooked into these state updates, so it has them defined as constants in https://github.com/OpenXT/toolstack-data/blob/master/dist/script/constants.js#L205.

I'd be fine with macro-ing them to something descriptive.

xenmgr/Vm/Monitor.hs
crogers1: I suppose so, I can change it so it's more readable.

richpersaud:
Please include JIRA ticket number.

rossphilipson:
Tested, working. Merging soon.

PR #78, opened 2017-04-05
[xenmgr] Reconnect vifs on ndvm reboot

crogers1:
Fix a few bugs along this control path in addition to new logic
  for applyVmBackendShift. When ndvm_x reboots, forall frontends
  with ndvm_x as a backend, cleanup all existing frontend nics and
  call into networkdaemon to hookup new vifs for ndvm_x.

  OXT-1011

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -897,7 +897,7 @@ bootVm config
 
       handleCreationPhases :: XM ()
       handleCreationPhases = do
-        waitForVmInternalState uuid CreatingDevices 30
+        waitForVmInternalState uuid CreatingDevices Running 30
 
         --Move these tasks up earlier in the guest boot process. Prevents the need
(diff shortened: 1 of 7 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/754

jean-edouard:
Code LGTM.
Some changes seem purely cosmetic, but make the code more readable, so +1.

rossphilipson:
So with these fixes in (and probably due to this one), when you restart the NDVM it starts endlessly creating vif nodes (I had wifi off). It is still going and Windows PnP is trying its best to keep up (it is still going as I speak) but I assume it will crash at some point. So not fixed...

crogers1:
@rossphilipson: I updated all the relevant PRs.  Instead of going through network daemon, I'm calling xl network-attach from xenmgr where I have all the information about the vif and can be very specific.  I also added 'devid' as a valid param to network-attach to prevent that endless vif creation loop you saw.  Worth rebuilding everything trying the test again.

PR #79, opened 2017-04-06
OXT-1067: Unfork: Use upstream Xen OCaml libraries

dozylynx:
Makefile updated with new (as of 2011) library names.
db_methods.ml : Xs -> Xenstore.Xs

Signed-off-by: Christopher Clark <christopher.clark6@baesystems.com>

### DIFF
--- a/dbd/Makefile
+++ b/dbd/Makefile
@@ -18,7 +18,7 @@
 
 include common.make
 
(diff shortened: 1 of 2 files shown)


rossphilipson:
LGTM

PR #81, opened 2017-04-20
[xenmgr] Fix path to remove snapshot on shutdown

crogers1:
OXT-1018

### DIFF
--- a/xenmgr/Vm/React.hs
+++ b/xenmgr/Vm/React.hs
@@ -285,9 +285,11 @@ maybeCleanupSnapshots = do
     return ()
   where
     removeIfExists path = do
-                            exists <- liftIO $ doesFileExist path
+                            let name = ("snap_"++) $ last $ split '/' path
+                            let path_actual = intercalate "/" ((init $ split '/' path) ++ [name])
+                            exists <- liftIO $ doesFileExist path_actual
                             case exists of
-                                True  -> do liftIO $ removeFile path
+                                True  -> do liftIO $ removeFile path_actual
                                             return ()
                                 False -> return ()
 


crogers1:
Now follows new naming scheme for snapshots so it can locate them for removal on vm shutdown.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/812

jean-edouard:
That snap_ prefix is weird, I wonder why/when it was added, since the file extensions already show it's a snapshot...
About your implementation of the fix, couldn't you just add the "snap_" prefix in the calling line? Something like:
`sequence $ map removeIfExists $ map ("snap_"++ . ++".snap.tmp.vhd") $ map diskPath disks`

crogers1:
The whole snapshot support (disk persistence) was implemented in xenvm before.  This commit implemented it in xenmgr, where it arguably should've been done in the first place: 
https://github.com/OpenXT/manager/commit/b6915e5828438037bed55dabc61bb09094e6e7ae

This commit implemented support for encrypted snapshots, which were never used before:
https://github.com/OpenXT/manager/commit/01da05782d731968b08fb3add43d49d570522445

The "snap_" prefix was added because the logic in the blktap encryption for key lookup is broken when the snapshot and parent disk share the same basename.  Rather than mess with our disk encryption code, I figured it was less intrusive to change the naming scheme on our snapshots.  Encrypted snapshots are a new feature anyway (compared to pre 7.0).

crogers1:
```
map (("snap_"++) . (++".snap.tmp.vhd")) $ map diskPath disks
```
That would add "snap_" to the front of the absolute path: "snap_/storage/disks/asdfasdfasdf.vhd".

jean-edouard:
Yeah you're right, ok, LGTM then!
Merging soon.

crogers1:
I'll need to open a PR against stable-7 for this too right?

jean-edouard:
The ticket has a Fix version of 7.0.0 so I guess you do.

PR #91, opened 2017-05-05
STABLE-7: Oxt 1091 stable7

crogers1:
isn't listening for acpi signals. Also write hvm-shutdown node
  that qemu acpi code is watching.

  OXT-1091


eric-ch:
To be merged with: https://github.com/OpenXT/xenclient-oe/pull/628

PR #92, opened 2017-05-08
OXT-1082: xenmgr MMIO hole size detection.

eric-ch:
XenVM used to adjust the MMIO hole of the guest depending on PCI
devices, and there resources, being passed-through. This behaviour is
now missing, leaving the MMIO hole too small to accomodate all the
required iomem resources below 4g.

OXT-1082

### DIFF
--- a/xenmgr/Vm/Config.hs
+++ b/xenmgr/Vm/Config.hs
@@ -105,6 +105,7 @@ import Tools.IfM
 import Rpc.Core
 import Vm.Types
 import Vm.Policies
+import Vm.Pci
 import XenMgr.Db
 import XenMgr.Host
 import XenMgr.Notify
@@ -694,8 +695,13 @@ pciSpecs :: VmConfig -> Rpc [PciSpec]
 pciSpecs cfg = do
     let devices = vmcfgPciPtDevices cfg
         uuid    = vmcfgUuid cfg
+        pciSysfsResFiles = map (++ "/resource")
+                           $ map (\(PciPtDev d _ _ _) -> pciSysfsDevPath $ devAddr d) devices
+    pciResources <- liftIO $ mapM pciGetMMIOResources pciSysfsResFiles
+    let mmioHoleAdjusted = 0x100000000 - (pciGetMemHoleBase . pciGetMemHole $ concat pciResources)
 
-    return $ ["pci=[" ++ (concat (intersperse "," (map (\dev -> "'" ++ stringAddr dev ++ "'") devices))) ++ "]"]
+    return $ [ "pci=[" ++ (concat (intersperse "," (map (\dev -> "'" ++ stringAddr dev ++ "'") devices))) ++ "]"
+             , "mmio_hole=" ++ (show . toMB $ mmioHoleAdjusted) ]
  where
    stringAddr (PciPtDev d _ _ _) =
            printf "%04x:%02x:%02x.%x"
@@ -704,6 +710,7 @@ pciSpecs cfg = do
                (pciSlot   addr)
                (pciFunc   addr)
        where addr = devAddr d
+   toMB size = div size (1024 * 1024)
 
 cpuidResponses :: VmConfig -> [String]
 cpuidResponses cfg = map option (vmcfgCpuidResponses cfg) where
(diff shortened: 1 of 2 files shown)


richpersaud:
Could this functionality be located in libxl, to benefit other libxl users of GPU pt?

eric-ch:
@richpersaud I was hesitant to add this to `libxl` with a patch, but it felt a bit redundant and unnecessary. This is populating the `mmio_hole` parameter of `xl` configuration which let users define arbitrary PCI hole sizes from 64MB to 2GB iirc. `xenmgr` being a higher level toolstack, it felt more appropriate to let it process the PCI hole size depending on what is passed-through when the VM boots. Actually it could be interesting to let `xenmgr` set this value from its configuration file like `xl` would and override the calculation as well (to leave room for another device to be assigned?)

richpersaud:
Having this in the configuration file might also be more toolstack-neutral, e.g. for future migration to libvirt.

eric-ch:
@richpersaud I am not sure to understand why providing an override would make this toolstack-neutral (or the opposite). I assume libvirt does something similar, and probably even has an argument to override that calculation and pass a specific `mmio_hole` to xl. I think a default calculation by the high-level toolstack (be it libvirt, xenmgr, or anything else) is preferable if the user does not specify the MMIO hole size, or even knows about it, to avoid VM failure.

richpersaud:
Sorry for being unclear in my previous comment:  I'm agreeing with default calculation & opportunity for manual override via config file.  If we do a default calculation and then save that into a config file, that value can be used after migration to another toolstack, even if their default calculation is different.

For toolstacks which support GPU PT (xapi, libvirt, xenmgr), I wonder if these calculations are much different in practice.

eric-ch:
@richpersaud Oh I see what you mean. `Save`, `Restore` and `Migration` are not possible with Xen when the guest has a PCI passed-through, at least without unplugging the passed-through PCI device ([afaict](https://wiki.xen.org/wiki/Xen_PCI_Passthrough#Xen_PCI_passthru_limitations)).
Moving the backend storage to another toolstack does not require the new toolstack to know about the MMIO hole size. If anything, it should re-calculate or use its own as it might pass a different device or have different requirements (more space for PCI pass-through hotplug, resources relocated above 4G limit?, ...)

eric-ch:
Build [834](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/834)

xenmgr/Vm/Config.hs
```diff
@@ -704,6 +710,7 @@ pciSpecs cfg = do
                (pciSlot   addr)
                (pciFunc   addr)
        where addr = devAddr d
+   toMB size = flip div (1024 * 1024) $ size
```
crogers1: flip() is unnecessary here.  Just switch the order of the arguments:
```haskell
div size (1024 * 1024)
```

xenmgr/Vm/Config.hs
```diff
@@ -694,8 +695,13 @@ pciSpecs :: VmConfig -> Rpc [PciSpec]
 pciSpecs cfg = do
     let devices = vmcfgPciPtDevices cfg
         uuid    = vmcfgUuid cfg
+        pciSysfsResFiles = map (flip (++) "/resource")
```
crogers1: Same with flip() here.  I'm assuming the goal is to prepend "/resource" to a list of sysfs paths?
```haskell
map ("/resource" ++)
```
should work fine.

eric-ch:
@crogers1 Removed the ```flip``` and switched the type of the resources to match 64bits integers presented in the files (used ```Data.Word``` as we discussed).

xenmgr/Vm/Pci.hs
```diff
@@ -449,7 +453,37 @@ initCache = PciCache <$> enumerateDevices
                  , TE.decodeUtf8 (deDeviceName e)]
       | otherwise = findName es id
 
-
+pciGetResources :: String -> IO ([(Word, Word, Word)])
```
crogers1: Could you comment this function with an example of what the input looks like that you're parsing from the resource file?  Would help make the function more readable for people looking at this down the road.

eric-ch: Done, also I renamed the function since it is only parsing MMIO resources.

crogers1:
Tested build 834, windows 7 guest 8gb RAM w/tools, passthrough Nvidia Quadro K2000, installed drivers, no crashes.  Boots into Windows just fine.

Haskell code looks good, @eric-ch if you squash those commits I'll give my +1.

crogers1:
+1, merging soon

PR #93, opened 2017-05-08
STABLE-7: OXT-1082: xenmgr MMIO hole size detection.

eric-ch:
XenVM used to adjust the MMIO hole of the guest depending on PCI
devices, and there resources, being passed-through. This behaviour is
now missing, leaving the MMIO hole too small to accomodate all the
required iomem resources below 4g.


crogers1:
Merging soon

PR #82, opened 2017-04-26
Migration to hash WPA-PSK passwords

troyc:
OXT-923 was addressed by causing nm-applet only store PSK passwords in hashed form, but any plaintext passwords in the dom-store will remain when OpenXT is upgraded. This PR adds a migration to hash all PSK passwords in the dom-store.

I used migration number 35 since the pending xl migration is using 34 (which I believe will also cause a conflict when merged).

PR #94, opened 2017-05-25
STABLE-7: OXT-923: migrations psk partial

eric-ch:
Original PR:
- https://github.com/OpenXT/manager/pull/82


rossphilipson:
@eric-ch tested this on master, looks good, will merge with RM sign off.

PR #97, opened 2017-08-17
[Xen 4.9] Increase stubdom memory from 80M to 96M

dozylynx:
80M has been observed as too low to allow a guest VM to reboot.

OXT-1203 for Xen upgrade.

Signed-off-by: Christopher Clark <christopher.clark6@baesystems.com>


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1078

PR #100, opened 2017-09-05
STABLE-7: [xenmgr] Don't use temporary-encrypted

crogers1:
The option isn't implemented well enough, just key off whether
  the encryption key is set or not to make an encrypted snapshot.

  OXT-966


jean-edouard:
@dpsmith please review.

jean-edouard:
Merging soon.

PR #101, opened 2017-09-06
[xenmgr] tap-ctl destroy supports d flag again

crogers1:
so use it to destroy the tap devices created during pv kernel
  extraction.

  OXT-1207

### DIFF
--- a/xenmgr/Vm/Utility.hs
+++ b/xenmgr/Vm/Utility.hs
@@ -74,18 +74,8 @@ tapCreateVhd = tapCreate "vhd"
 tapCreateVdi = tapCreate "vdi"
 tapCreateAio = tapCreate "aio"
 
---tapDestroy syntax changed as well, requires more information than just the tap device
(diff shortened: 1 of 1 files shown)


crogers1:
When I did the xl merge, the tap-ctl destroy -d wasn't a valid option, so the code you see in the (-)diff was an approach to address that.  Turns out that code is buggy anyway, so now that tap-ctl has -d again due to some version uprev we made, let's just go back to using it.

For reference, the path that is passed to -d is the device path (/dev/xen/blktap-2/tapdevX)

jean-edouard:
LGTM, building soon.

jean-edouard:
Actually, @crogers1, couldn't you use `readProcessOrDie` instead of `spawnShell'`? `spawnShell'` has the undesirable effect of leaking file descriptors.

crogers1:
Oh I didn't realize that, yea I can swap that quick

jean-edouard:
LGTM, merging soon.

PR #105, opened 2017-09-21
[xenmgr] Hook up passthrough-mmio/io for xl

crogers1:
passthrough-mmio and passthrough-io map to iomem and ioports
  in xl.  Hook up the logic in Config.hs to dump them into the config

  OXT-1220

### DIFF
--- a/xenmgr/Vm/Config.hs
+++ b/xenmgr/Vm/Config.hs
@@ -835,8 +835,10 @@ miscSpecs cfg = do
           , ("seclabel"        , vmFlaskLabel)
           , ("serial"          , vmSerial)
           , ("stubdom_cmdline" , vmStubdomCmdline)
-          , ("stubdom_memory"  , vmStubdomMemory)
-          ]
+          , ("stubdom_memory"  , vmStubdomMemory)   --OXT-1220: iomem and ioports should be reworked to support specifying multiple
+          , ("iomem"           , vmPassthroughMmio) --ranges at a finer granularity. Few ways to implement, likely as a db-node with
+          , ("ioports"         , vmPassthroughIo)   --each range as an entry beneath it, which is read and parsed during xl cfg generation.
+          ]                                         --Remove this comment block when implemented.
 
       -- xl config handles certain options different than others (eg. quotes, brackets)
       -- we format them on a case by case basis here before sending them off to xl.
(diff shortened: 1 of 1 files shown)


crogers1:
https://github.com/OpenXT/manager/pull/106 -> stable-7

Built ok, needs testing (not currently near my test box).  Use xec-vm to set these fields, use the xl syntax, comma separated per iomem/ioport entry.  Xenmgr will take care of wrapping them in quotes and brackets when it generates the xl domain config in /tmp on vm start.

xenmgr/Vm/Config.hs
```diff
@@ -852,6 +854,8 @@ miscSpecs cfg = do
                                 _       -> case name of
                                              "viridian" -> name ++ "=" ++ (wrapBrackets $ wrapQuotes v)
                                              "serial"   -> name ++ "=" ++ (wrapBrackets $ wrapQuotes v)
+                                             "iomem"   -> name ++ "=" ++ (wrapBrackets $ wrapQuotes v)
```
jandryuk: This naively only supports a single value when iomem can handle multiple ranges:
iomem=[ "IOMEM_START,NUM_PAGES[@GFN]", "IOMEM_START,NUM_PAGES[@GFN]", ... ]

You can jam multiple values in by writing something like "0xfed40,4','0xfff60,1" so the inner single quotes are passed through and interpreted by xl.

richpersaud:
I don't think anyone is currently using this feature in production, so we can drop backwards compatibility in order to simplify support for XL syntax.  This feature is useful for demo purposes.

eric-ch:
Building [1164](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1164)

jandryuk:
I recommend re-working this to accept multiple range definitions without resorting to cramming in partially quoted strings.

jandryuk:
Works.

Created an HVM CentOS 7 guest.

Passthrough the TPM mmio ranges:
`db-write /vm/$( xec-vm CentOS7 get uuid )/config/passthrough-mmio "fed40,5"`
modify the CentOS7 VM /boot/grub2/grub.cfg appending "tpm_tis.force=1" to the linux command line.
Boot/reboot.

`cat /sys/class/tpm/tpm0/device/pcrs` in the guest and see the PCRs from the TPM.

crogers1:
PR updated with in-code comment block with ticket id and info regarding future implementation where ranges are added/removed at a finer granularity.

PR #106, opened 2017-09-21
STABLE-7: [xenmgr] Hook up passthrough-mmio/io for xl

crogers1:
passthrough-mmio and passthrough-io map to iomem and ioports
  in xl.  Hook up the logic in Config.hs to dump them into the config

  OXT-1220


crogers1:
See my latest post on: https://github.com/OpenXT/manager/pull/105
regarding the added in-code comment block.

PR #107, opened 2017-12-14
wait for domain to shutdown

mahantesh-ais:
OXT-1221

### DIFF
--- a/apptool/Import/Images.hs
+++ b/apptool/Import/Images.hs
@@ -306,7 +306,7 @@ finally' = flip E.finally
 withCryptoVhdTap :: FilePath -> (FilePath -> IO a) -> Import a
 withCryptoVhdTap vhdfile action = do
   keys <- tempFileName =<< cryptoKeysFolder
-  liftIO $ withVhdTap [("TAPDISK2_CRYPTO_KEYDIR", keys)] vhdfile action
+  liftIO $ withVhdTap [("TAPDISK2_CRYPTO_KEYDIR", keys), ("TAPDISK3_CRYPTO_KEYDIR", keys)] vhdfile action
 
 withVhdTap :: [(String,String)] -> FilePath -> (FilePath -> IO a) -> IO a
 withVhdTap extraenv vhdfile action = E.bracket (tapCreateVhd extraenv vhdfile) tapDestroy action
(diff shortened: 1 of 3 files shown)


dpsmith:
I think this need to be considered carefully and well exercised with good coverage testing. I don't know what threading model(if there is one at all) is in place, but I believe this will freeze the processing thread that handles this logic until the target domain shuts down. If xenmgr is single threaded or uses green threads, then this would mean xenmgr would likely be completely unresponsive.

jean-edouard:
This is actually fixing a regression introduced by the xl support. The pre-xl code was blocking on shutdown, and I never saw xenmgr freeze.

dpsmith:
But before xl, domain shutdown was handled in xenvm. Are you sure that xenmgr did a synchronous wait on xenvm or for did it listen for [notify signal](https://github.com/OpenXT/idl/blob/master/interfaces/xenvm.xml#L5) event from xenvm?

jean-edouard:
@crogers1 do you have more details about this?
Thanks

richpersaud:
This PR has implications that extend beyond blktap3, so it would require a separate JIRA ticket, which can be a dependency of OXT-1221.  Discussions of LibXL and xenmgr state machines can be covered in this JIRA ticket, cross-referencing other JIRA tickets which impacted state management, e.g. clearly identifying any previous JIRA tickets which affect shutdown state.

crogers1:
The haskell function
```system:: String -> IO ExitCode```
is already blocking, so adding -w will indeed wait for the xl process to terminate before finishing the shutdown call in xenmgr.  @dpsmith, incoming RPC (so dbus, or ```xec-vm```) calls are forked into a lightweight IO thread and handled to completion.  This shouldn't block subsequent RPC calls into xenmgr from other sources.  You can see this code in ```handleIncoming``` in xclibs/xch-rpc/Rpc/Core.hs.

@richpersaud, this code change seems like it's a required component of OXT-1221 and its impact is not as far ranging as one might think, indicated in my above response to @dpsmith.  I would argue a separate JIRA ticket is probably out of scope.

The one thing I would absolutely test for is to make sure the xenmgr callbacks still fire when the xl process sets the domain state to "shutdown" in the xenstore.  I would expect them to, since the RPC handling thread is separate from the vm's monitoring thread, and the state update callback fires within the vm's monitoring thread.  But we can never be too careful.

jean-edouard:
LGTM, merging around end of day.

PR #113, opened 2018-04-25
Oxt 1299

crogers1:
OXT-1299


crogers1:
Merge with https://github.com/OpenXT/xenclient-oe/pull/877

PR #114, opened 2018-04-25
[manager] Fixup domain reboot without sleeps

crogers1:
OXT-1299

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -23,6 +23,7 @@ module Vm.Actions
           , trashUnusedServiceVms
           , createVm, CreateVmPms(..), defaultCreateVmPms
           , removeVm
+          , restartVm
           , startVm
           , startVmInternal
           , rebootVm
@@ -227,7 +228,7 @@ startServiceVm uuid = xmContext >>= \xm -> liftRpc $
                          else liftIO $ do
                            xsWrite (vmSuspendImageStatePath uuid) "start-no-snapshot"
 
-                      runXM xm (startVm False uuid)
+                      runXM xm (startVm uuid)
           True  -> info $ "service vm " ++ show uuid ++ " already running"
     where
       snapshot_request xm file =
@@ -243,7 +244,7 @@ startServiceVm uuid = xmContext >>= \xm -> liftRpc $
                   info $ "DONE taking memory image snapshot for service vm " ++ show uuid
                   liftIO $ xsWrite (vmSuspendImageStatePath uuid) "snapshot-done"
                   -- double start, TODO: maybe wont be necessary
-                  runXM xm (startVm False uuid)
+                  runXM xm (startVm uuid)
                   -- finished waiting on this watch
                   return True
 
@@ -479,8 +480,14 @@ getVhdReferences vhd = concat <$> (mapM (diskVhdReferences vhd) =<< getVms) wher
   diskVhdReferences vhd vm = zip (repeat vm) . filter (references vhd) . M.elems <$> getDisks vm where
     references vhd disk = diskPath disk == vhd
 
-startVm :: Bool -> Uuid -> XM ()
-startVm is_reboot uuid = do
+startVm :: Uuid -> XM ()
+startVm uuid = _startVm False uuid
+
+restartVm :: Uuid -> XM ()
+restartVm uuid = _startVm True uuid
+
+_startVm :: Bool -> Uuid -> XM ()
+_startVm is_reboot uuid = do
   withPreCreationState uuid $ do
     ran <- liftRpc $ runEventScript HardFail uuid getVmRunInsteadofStart [uuidStr uuid]
     when (not ran) $ startVmInternal uuid is_reboot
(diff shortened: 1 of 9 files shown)


crogers1:
Merge with https://github.com/OpenXT/xenclient-oe/pull/877

xenmgr/Vm/Actions.hs
```diff
@@ -479,11 +479,11 @@ getVhdReferences vhd = concat <$> (mapM (diskVhdReferences vhd) =<< getVms) wher
   diskVhdReferences vhd vm = zip (repeat vm) . filter (references vhd) . M.elems <$> getDisks vm where
     references vhd disk = diskPath disk == vhd
 
-startVm :: Uuid -> XM ()
-startVm uuid = do
+startVm :: Bool -> Uuid -> XM ()
+startVm is_reboot uuid = do
```
jandryuk: What do you think about doing:
```
startVm uuid = do _startVM False uuid
restartVm uuid = do _startVm True uuid
_startVm is_reboot uuid = do...
```
That way the assorted startVm False and startVm True are replaced by startVm and restartVm and you don't have to know what the boolean argument means.

xenmgr/XenMgr/Connect/Xl.hs
```diff
@@ -201,22 +206,51 @@ unpause uuid = do
                 exitCode <- system ("xl unpause " ++ domid)
                 bailIfError exitCode "Error unpausing domain."
 
+getXlProcess :: Uuid -> IO String
+getXlProcess uuid = do
+    let conv_uuid = TT.replace "_" "-" (show uuid)
+    (ec,str_pid,_) <- readProcessWithExitCode "pgrep" ["-f", conv_uuid] ""
```
jandryuk: Maybe just save the xl pid when it is originally forked?

xenmgr/Vm/Actions.hs
crogers1: Yea wrapping that startVm call is probably the way to go to make this more readable.  I like it.

xenmgr/XenMgr/Connect/Xl.hs
jandryuk: Well it's not so much the expense, but more that xenmgr should keep track of its children.  However xl daemonizes, so it's not as simple as just tracking the spawned child.

Um, this just matches off the uuid, but there is no verification that it is an xl process.

crogers1: Good point, xl processes were the only ones I found on the platform that ever have the full uuid, but that may not be the case forever.

crogers1:
Updated with @jandryuk's recommendations

xenmgr/XenMgr/Connect/Xl.hs
```diff
@@ -209,7 +209,7 @@ unpause uuid = do
 getXlProcess :: Uuid -> IO String
 getXlProcess uuid = do
     let conv_uuid = TT.replace "_" "-" (show uuid)
-    (ec,str_pid,_) <- readProcessWithExitCode "pgrep" ["-f", conv_uuid] ""
+    (ec,str_pid,_) <- readProcessWithExitCode "pgrep" ["-f", "xl[a-zA-Z0-9\\w:\\.\\?\\+\\-]*" ++ conv_uuid] ""
```
jandryuk: Minimally, a more robust check could start with "^xl " to ensure the match begins with xl and have a space to separate the command from arguments.  xenmgr starts the xl process with ```createProcess (proc "xl" ["create", configPath uuid, "-p"])``` so the expression could just be ```"^xl create " ++ configPath uuid ++ " -p"```?

Since we'll signaling processes, I still think it would be better to track pids explicitly and ensure we are only signaling processes xenmgr started.

crogers1: You're right about the expression to look for, I don't need that match set.  Will update.

You mentioned before the difficulty of tracking pids due to xl forking.  Xenmgr can know the pid from the initial "xl create" call, but getting the forked pid would either require a pgrep in some form anyway, or a message from xl -> xenmgr telling it the pid changed, which is something I'm trying to avoid.  There's only two things I want going from xl -> xenmgr: errors on failure, and domain state changes (which is more indirectly sent through xenstore).

jandryuk: Yes, but you can use "-F" to run in the foreground and skip daemonizing.  With that the pid wouldn't change from the initial exec.  I don't know if that would then require further changes to work properly like waiting on children.

I was just re-stating my preference.  With the pgrep improvements this will work.  A future improvement could be to change over to tracking children.

crogers1:
Updated with more specific regex for pgrep

eric-ch:
Build [$1498](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1498)

jean-edouard:
Build here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1531

jean-edouard:
Merging soon.

PR #117, opened 2018-06-07
xec-vm: fix suspend/resume to/from file

Jafaral:
With suspend and resume, xec-vm fails with the error:

   unable to retrieve domain configuration

Supplying the extra command parameter (the domain's xen config file)
to the underlying xl command fixes the issue.

Signed-off-by: Jafar Al-Gharaibeh <jafar@atcorp.com>

### DIFF
--- a/xenmgr/XenMgr/Connect/Xl.hs
+++ b/xenmgr/XenMgr/Connect/Xl.hs
@@ -286,14 +286,14 @@ suspendToFile :: Uuid -> FilePath -> IO ()
 suspendToFile uuid file =
     do
       domid    <- getDomainId uuid
-      exitCode <- system ("xl save " ++ domid ++ " " ++ file)
+      exitCode <- system ("xl save " ++ domid ++ " " ++ file ++ " " ++ configPath uuid)
       bailIfError exitCode "Error suspending to file."
 
 resumeFromFile :: Uuid -> FilePath -> Bool -> Bool -> IO ()
 resumeFromFile uuid file delete paused =
     do
       let p = if paused then "-p" else ""
-      _ <- system ("xl restore " ++ p ++ file)
+      _ <- system ("xl restore " ++ p ++ " " ++ configPath uuid ++ " " ++ file)
       if delete then removeFile file else return ()
 
 --Ask xl directly for the domid


jean-edouard:
That fails, pretty sure you're missing a space here...

Jafaral:
@jean-edouard where exactly ?

xenmgr/XenMgr/Connect/Xl.hs
```diff
@@ -286,14 +286,14 @@ suspendToFile :: Uuid -> FilePath -> IO ()
 suspendToFile uuid file =
     do
       domid    <- getDomainId uuid
-      exitCode <- system ("xl save " ++ domid ++ " " ++ file)
+      exitCode <- system ("xl save " ++ domid ++ " " ++ file ++ configPath uuid)
```
eric-ch: `s/file ++ configPath uuid/file ++ " " ++ configPath uuid/`

xenmgr/XenMgr/Connect/Xl.hs
```diff
@@ -286,14 +286,14 @@ suspendToFile :: Uuid -> FilePath -> IO ()
...
       bailIfError exitCode "Error suspending to file."
 
 resumeFromFile :: Uuid -> FilePath -> Bool -> Bool -> IO ()
 resumeFromFile uuid file delete paused =
     do
       let p = if paused then "-p" else ""
-      _ <- system ("xl restore " ++ p ++ file)
+      _ <- system ("xl restore " ++ p ++ configPath uuid ++ file)
```
eric-ch: `s/p ++ configPath uuid ++ file/p ++ " " ++ configPath uuid ++ " " ++ file/`

Jafaral:
Thanks @eric-ch , I was thinking about this patch this morning and I want to change it anyway. I think the call to configPath needs to be pulled out of the system command. If the user manually deletes the file, or the file gets deleted for any reason configPath would fail. I will do that and update the PR after I test. I will make sure to literally copy the code from my dev machine this time instead of replicating it. :)

Jafaral:
Interesting that the original code didn't have the space either in "restore" case!

eric-ch:
I believe several other `system()` issued commands already rely on the file existing and being present (otherwise the command would only fail, logging the error output?). If you want to take a look at that, we can already add this PR (albeit, with the spaces) so people can already try out the save/restore commands.

> the original code didn't have the space either

Yes, it seems that was broken :), or I am missing something...

Jafaral:
Ok then. I will fix the spaces and update the PR.

Jafaral:
@jean-edouard @eric-ch  PR updated!

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1562

eric-ch:
Can you PR against stable-8 as well please?
Without these changes, the commands are broken.

jean-edouard:
I'll have a look at the XSM policy bits for this tomorrow.

Jafaral:
Just pushed a PR against stable-8.
@jean-edouard  that would be great. I have been working on my test machine with XSM disabled so I didn't have to worry about it. I didn't look at XSM policy before, so this would be a good chance for me to learn.

jean-edouard:
Still failing... I'm using a 64 bits Windows 10 guest with tools, with XSM and SELinux both in permissive mode.

Jafaral:
here is a run my latest build, ad I still get:

xec: invoked with: -o /vm/668935d2_92b8_4728_adc3_059661441ba5 suspend-to-file hi.save

as the underlying command. Obviously, it is still missing the xen xonfig file argument which is what this patch is supposed to do. Either my latest build missed that patch, or maybe there I'm missing something from the full picture here.  It seems that Xl.hs is the only place where I need to make the a change.

jean-edouard:
Yeah I get the same result with build $1562

Jafaral:
@eric-ch I was talking about this. 

Do you have an idea of how the underlying xec command is not getting the config file in my case and also Jed's case ?

Jafaral:
I looked at the code again, configPath uuid should never fail, so we always should have that argument even if the actual config files doesn't exist.

Jafaral:
BTW,  Are there any unit tests for  things like Xl.hs.  My experience with Haskell is close to nil, it helps if there is a way to tell the code is doing what I think it does.

jean-edouard:
Also the Xen logs seem to show unhappy Windows drivers:
```
(d5)  XENUTIL: WARNING: SuspendThread: pending tokens:
(d5)  XENUTIL: WARNING: SuspendThread: xenbus probe
(d5)  XENUTIL: WARNING: SuspendThread: xenbus probe
(d5)  XENUTIL: WARNING: SuspendThread: xenbus probe
[...]
```

Jafaral:
@jean-edouard, right, I was looking at "xec" whereas the underlying command is "xl". 

In my initial testing invoking xl directly from the command line worked when stub-domains were disabled. It failed if stud-domains are enabled. And yes.

As I mentioned on
https://openxt.atlassian.net/browse/OXT-1289?filter=-5&jql=resolution%20%3D%20Unresolved%20ORDER%20BY%20created%20DESC%2C%20priority%20DESC%2C%20updated%20DESC

I realized that this requires fixes/changes in many places. I have been tackling the issues/bugs one at a time. Since you confirmed xl now actually gets the correct arguments, I will look into why invoking "xec" still doesn't work.

jean-edouard:
This PR and https://github.com/OpenXT/xenclient-oe/pull/916 don't fix hibernation in OpenXT, but are a good start.
I will merge them tomorrow if there's no objection.

PR #119, opened 2018-06-13
STABLE-8: xec-vm: fix suspend/resume to/from file

Jafaral:
With suspend and resume, xec-vm fails with the error:

       unable to retrieve domain configuration

    Supplying the extra command parameter (the domain's xen config file)
    to the underlying xl command fixes the issue.

Signed-off-by: Jafar Al-Gharaibeh <jafar@atcorp.com>

### DIFF
--- a/xenmgr/XenMgr/Connect/Xl.hs
+++ b/xenmgr/XenMgr/Connect/Xl.hs
@@ -251,14 +251,14 @@ suspendToFile :: Uuid -> FilePath -> IO ()
 suspendToFile uuid file =
     do
       domid    <- getDomainId uuid
-      exitCode <- system ("xl save " ++ domid ++ " " ++ file)
+      exitCode <- system ("xl save " ++ domid ++ " " ++ file ++ " " ++ configPath uuid)
       bailIfError exitCode "Error suspending to file."
 
 resumeFromFile :: Uuid -> FilePath -> Bool -> Bool -> IO ()
 resumeFromFile uuid file delete paused =
     do
       let p = if paused then "-p" else ""
-      _ <- system ("xl restore " ++ p ++ file)
+      _ <- system ("xl restore " ++ p ++ " " ++ configPath uuid ++ " " ++ file)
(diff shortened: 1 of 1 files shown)


eric-ch:
@Jafaral No sure what the context is regarding your last comment...

This PR is not working in master. You can ignore or close this one until the master submission is fixed if that makes things easier?

Also, I presume "fetch" refers to [fetch.sh](https://github.com/OpenXT/openxt/blob/master/build-scripts/fetch.sh). This only `git fetch` the mirrors created by the build-scripts.
Openembedded has its own git mirrors, that it uses to keep actual clones in the `TMPDIR` (where the sources are checked out).

Jafaral:
@eric-ch , I was wondering where I posted this comment  :D 
It was supposed to be for the PR on master. I was generally speaking about having to remove some mirrors on the build machine (/home/git/openxt) do git clone --bare for my own repos, and those don't seem to get updated form my own github repos. Anyway, I just have to learn to streamline some of those ruff edges.

jean-edouard:
For `git fetch` to update bare git repos, those have to be mirrors (see `git clone --mirror`, or https://github.com/OpenXT/openxt/blob/master/build-scripts/setup.sh#L291)

jean-edouard:
By the way, still unrelated to this PR (sorry), but feedback on the openxt.git build-scripts is very welcome.
If you have any improvement idea, please file a Jira ticket.

Jafaral:
Thanks @jean-edouard . I will report any issue or let you know if I have ideas.

jean-edouard:
This PR is part of a larger effort.
I don't think we should merge it until the whole feature is fixed in master.
@Jafaral, I'd recommend closing it for now, it can always be reopened later.
@eric-ch, I don't think we should plan on supporting hibernation in 8.0.0.

Jafaral:
@jean-edouard , I agree. this doesn't need to go in 8.0.0 as this PR alone isn't enough to get suspend/resume to work. I will close it for now.

PR #122, opened 2018-06-21
STABLE-8: Fix XL syntax for raw disks

jean-edouard:
(cherry picked from commit d6e1121b05289180db0e06d8f629938a425aa04b)

### DIFF
--- a/xenmgr/Vm/Config.hs
+++ b/xenmgr/Vm/Config.hs
@@ -629,8 +629,8 @@ diskSpec uuid d  = do
     cdType stubdom d =
       case (enumMarshall $ diskDeviceType d) of
           "cdrom" -> if stubdom then "backendtype=tap" else "backendtype=phy"
-          _       -> "backendtype=tap"
-    fileToRaw typ = if typ == "file" then "raw" else typ
+          _       -> if (enumMarshall $ diskType d) == "phy" then "backendtype=phy" else "backendtype=tap"
+    fileToRaw typ = if typ == "file" || typ == "phy" then "raw" else typ
 
 -- Next section: information about Network Interfaces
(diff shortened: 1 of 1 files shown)


eric-ch:
Will merge soon.

rneilturner:
Although not blocking this PR, there is still an issue with using raw disks inside of a stubdom. Also, our use case doesn't allow for non-stubdoms so we never tested the raw disk access outside of a stubdom. We did validate this functionality worked with stubdom's turned on for stable-6. We will be investigating further what is needed for raw disk access on stable-8 with stubdom support.

eric-ch:
@rneilturner Interesting. I tried following [OXT-1356](https://openxt.atlassian.net/browse/OXT-1356) verification steps, which only check if the toolstack passes the guest starts and can detect the disk, but not its usage. What is the issue you are seeing?

rneilturner:
I will get the logs tomorrow from the test engineer. I believe it got as far as booting the stubdom and then received an error. He spent most of the afternoon validating stable-6 which worked fine. I don't think he was able to get very far with the triage.

eric-ch:
@rneilturner Any development on that?

deepin125:
This is what I was seeing when attempting to do the disk pass through.
[failure.txt](https://github.com/OpenXT/manager/files/2146792/failure.txt)

eric-ch:
Build: [$1590](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1590)

eric-ch:
@deepin125 Build: [$1590](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1590) repro the issue consistently for VMs with stub-domain. I'll comment the ticket so we can merge this first and investigate the stub-domain case.

PR #123, opened 2018-07-19
OXT-1384: Syncvm Flask label

jandryuk:
The current syncvm template creates an un-bootable config. The old template passed the flask label directly to xenm through extra-xenvm "flask-label". This worked for the xenvm-based system, but it broke with the XL conversion. "flask-label" is invalid in the XL config ("seclabel" should be used), so it throws an error about the legacy Python syntax.

The migration code hardcodes the syncvm_t flask label.  Alternate labels will be clobbered, but this doesn't seem like a realistic case to handle.

### DIFF
--- a/templates/default/new-vm-sync
+++ b/templates/default/new-vm-sync
@@ -25,9 +25,7 @@
     "display": "none",
     "cmdline": "root=\/dev\/xvda1 xencons=hvc0",
     "kernel-extract": "\/boot\/bzImage",
-    "extra-xenvm": {
-      "0": "flask-label=system_u:system_r:syncvm_t"
-    },
+    "flask-label": "system_u:system_r:syncvm_t",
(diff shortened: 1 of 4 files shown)


jean-edouard:
LGTM. This probably makes sense in stable-8 too.
Merging Monday.

eric-ch:
@jandryuk Can you PR against stable-8 as well please?

jandryuk:
HVM NDVM introduced Migration 37 into master which is not present in stable-8.  How do people feel about skipping Migration 37 in stable-8 to better keep stable-8 & master in sync for the time being?

jean-edouard:
I vote for swapping them. (move HVM NDVM to 38 and have this PR be 37)

eric-ch:
Will merge soon.

PR #124, opened 2018-07-24
STABLE-8: OXT-1384: Syncvm Flask label

jandryuk:
This is the stable-8 version of https://github.com/OpenXT/manager/pull/123

The current syncvm template creates an un-bootable config. The old template passed the flask label directly to xenvm through extra-xenvm "flask-label". This worked for the xenvm-based system, but it broke with the XL conversion. "flask-label" is invalid in the XL config ("seclabel" should be used), so it throws an error about the legacy Python syntax.

The migration code hardcodes the syncvm_t flask label. Alternate labels will be clobbered, but this doesn't seem like a realistic case to handle.

This stable-8 version skips db version 38.  Migrations/M_37.hs is a placeholder empty file.  In the log output you see `update-db: Migrating 37 -> 39` skipping the intermediate version.

### DIFF
--- a/templates/default/new-vm-sync
+++ b/templates/default/new-vm-sync
@@ -25,9 +25,7 @@
     "display": "none",
     "cmdline": "root=\/dev\/xvda1 xencons=hvc0",
     "kernel-extract": "\/boot\/bzImage",
-    "extra-xenvm": {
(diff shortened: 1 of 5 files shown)


eric-ch:
Will merge soon.

PR #125, opened 2018-08-01
OXT-1383: PVH support

jandryuk:
Switch from an HVM boolean to the new virt-type enum.

I'm not particularly thrilled with the migration from hvm to virt-type.  [Here](https://github.com/OpenXT/manager/commit/69aafcd08b2a354e13e11544ec48ff6259f7ba49#diff-63e043d740a20cd98bba292e85c5f188R40) I couldn't get `s` as a JSBool.  When I tried:
```
    Just s | s == JSBool True  -> makeHvm tree
    Just s | s == JSBool False -> makePv  tree
```
it seemed to take the makeHvm path for both true and false.  The existing code does work though.

[xenmgr: Convert NIC checks from stubdom to HVM](https://github.com/OpenXT/manager/commit/1f92cecedce973ff86f04501ce0ca29a3677c415) is a change, but I think it's valid.

Depends on: https://github.com/OpenXT/idl/pull/30


jandryuk:
PVH requires an XSM label that is a cross between PV and HVM.  Neither hvm_guest_t nor pv_guest_t can successfully boot PVH.  For my testing, I've changed the policy to have xen_pv_type(foo) and xen_hvm_type(foo).

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1653

xenmgr/Vm/Config.hs
```diff
@@ -686,9 +700,9 @@ nicSpec cfg amt eth0Mac nic networkDomID =
                 | Just mac <- eth0Mac, amt == True  = ["mac=" ++ unswizzleMac mac]
       -- Otherwise we do not touch the VM mac and let xenvm choose
                 | otherwise                         = [ ]
-      nicType   = if (vmcfgStubdom cfg) then ["type=ioemu"] else ["type=vif"]
-      modelType | Just model <- nicdefModel nic = if (vmcfgStubdom cfg) then ["model="++model] else []
-                | otherwise                     = if (vmcfgStubdom cfg) then ["model=e1000"] else []
+      nicType   = if isHvm cfg then ["type=ioemu"] else ["type=vif"]
+      modelType | Just model <- nicdefModel nic = if isHvm cfg then ["model="++model] else []
+                | otherwise                     = if isHvm cfg then ["model=e1000"] else []
```
jean-edouard: This assumes all HVMs have a stubdom, which is not true. It breaks the stubdom-less case by added emulated NICs to stubdom-less guests, causing hangs.

jandryuk: Okay, I can drop 1f92cecedce973ff86f04501ce0ca29a3677c415 which introduced this change.  I wasn't totally sure about it.  So an stubdom-less HVM will not get presented an emulated NIC at all since QEMU in Dom0 cannot network it properly.  But a stubdom-less HVM can still use PV drivers to NDVM.

What "hangs" are you seeing?  Hangs for actual networking, or something else?

Thanks for taking a look at this.

jandryuk:
Dropped the patch messing with the NIC.  Also deleted some dead code from the first patch.

jandryuk:
If you want to wait for @crogers1 to review, I am fine with that.  Especially if there is a nicer way to handle the migration in: https://github.com/OpenXT/manager/pull/125/commits/fae5991ad80ef74db7f0c85da021313a934fce10#diff-63e043d740a20cd98bba292e85c5f188R40

xenmgr/Vm/Config.hs
jean-edouard: It'd be fine to overload this `if` to handle the PVH case, but here it was completely repurposed :)
When I start a Win10 guest with tools and without a stubdom, it hangs on the Welcome screen, probably waiting for the emulated NICs to show up or something.

jandryuk: Yes, you're right. I'll have to redo it to check both hvm and stubdom.

jandryuk:
I re-pushed, adding https://github.com/OpenXT/manager/pull/125/commits/671b1bf6c28dcc1b38099b9bbf9312f91c4c2de8 to check for both HVM and stubdom before adding emulated NICs.

crogers1:
How does XL know when it's booting a PVH guest?  Just the absence of device_model_override?

jandryuk:
xl.cfg "type" field. https://xenbits.xen.org/docs/unstable/man/xl.cfg.5.html#Selecting-Guest-Type

It's written in this patch:
https://github.com/OpenXT/manager/commit/a095b59690360981369460c308c6fce097865cc0

crogers1:
Ah, I skipped right over that.  The PR looks good to me.

jandryuk:
Thanks, @crogers1 .  Do you have any thoughts on improving the migration: https://github.com/OpenXT/manager/pull/125/commits/5ac4c13037656057ac65341fe840d76a6ee2652c#diff-63e043d740a20cd98bba292e85c5f188R40

It's fragile since it relies on dbd storing booleans as the strings "true" and "false".  When I tried:

    Just s | s == JSBool True  -> makeHvm tree
    Just s | s == JSBool False -> makePv  tree

it seemed to take the makeHvm path for both true and false. The existing code does work though.  Looking at this again, is `Just s` on the second line wrong?

crogers1:
Yea so guards, ```|``` are just syntactic sugar for if-else, so you have two identical cases there.  You probably want
```case s of
      Just s -> s == JSBool True   | makeHvm tree
                     s == JSBool False | makePv tree
      Nothing -> (some sort of no-op here)
```
Since ```s``` is a Maybe we might be able to do a ```fromMaybe``` to get the ```JSBool``` with a default value if it's a ```Nothing```.

jandryuk:
Your suggestion fails to compile with:
> Migrations/M_39.hs:40:32: parse error on input `|'

`jsGet` is returning a `JSValue`, but there isn't a good way to change it into a `JSBool`.  There is `readJSBool :: GetJSON JSValue`, but we can't just pass a string into that - we don't seem to have `GetJSON`.

Eventually, I came up with this, which seems to work:
```
Just s  -> case decode $ jsUnboxString s of
             Ok True  -> makeHvm tree
             Ok False -> makePv  tree
             _        ->         tree
Nothing    -> tree
```
...but that doesn't seem like much of an improvement.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1735

PR #126, opened 2018-09-06
OXT-1431: Guest UEFI support

tklengyel:
BIOS option allows a user to choose between a guest using UEFI through OVMF, or traditional BIOS through SeaBios


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1739

PR #127, opened 2018-09-20
OXT-1442: xenmgr: Use xl option rtc_timeoffset

jandryuk:
time-offset is not a valid XL option - XL thinks it is python.  A
similar XL option is rtc_timeoffset which sets the real time clock
offset in seconds.  I don't know what the old units were for this option
though.  xenvm did not seem to do anything with the value.

OXT-1442


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1745

jean-edouard:
LGTM, merging soon.

PR #128, opened 2018-10-02
Stubdom/PVH fixes

jandryuk:
If a PVH domain has stubdom set to true, we hang waiting for a stubdom to appear.  Avoid this by making getVmStubdom only return true for HVMs.

The second commit removes some dead code.


eric-ch:
Building [$1772](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1772)

eric-ch:
Will merge soon.

PR #130, opened 2018-11-05
OXT-796: xenmgr: Match multi-strings in dmi chassis type.

eric-ch:
dmidecode may return more than one type, e.g:
```sh
root@e7470:~# dmidecode -s chassis-type
Laptop
Docking Station
```

A docked laptop is still a laptop.
Use a list of keywords and check every line for a match.


eric-ch:
v2:
- `elem` has an infix notation, use it instead of `flip`.

jean-edouard:
LGTM, please PR against stable-8 as well.

jean-edouard:
LGTM, merging soon.

PR #133, opened 2019-03-08
Add  AHCI support (Qemu SATA disk emulation)

mahantesh-ais:
OXT-1123

### DIFF
--- a/templates/default/new-vm
+++ b/templates/default/new-vm
@@ -26,6 +26,7 @@
     "display": "none",
     "boot": "cd",
     "flask-label": "system_u:system_r:hvm_guest_t",
+    "hdtype": "ahci",
     "disk": {
       "0": {
         "path": "\/storage\/isos\/xc-tools.iso",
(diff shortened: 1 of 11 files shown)


mahantesh-ais:
To be merged along with:
https://github.com/OpenXT/xenclient-oe/pull/1098
https://github.com/OpenXT/idl/pull/32

pearsonk:
Slight nitpick of naming, but why the selection of "hdtype"? This seems to imply hard drive, but the interface for SATA is beyond hard drives (ISOs). Would something more like "bustype" be a better representation?

mahantesh-ais:
I agree, but I carried the same name provided by xen(xl) config,
which can be seen here: http://xenbits.xen.org/docs/4.11-testing/man/xl.cfg.5.html

pearsonk:
Makes sense to me. Again, was just a nitpick. Thanks for taking this on, I've been wondering why we were only using IDE emulation for a while, especially with the limitations of VHDs, default virtual ISO, and ATAPI-pt.

mahantesh-ais:
Built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1896

jean-edouard:
Merging soon.

PR #134, opened 2019-03-13
x86_64: Xenmgr 64-bit support

crogers1:


crogers1:
See https://github.com/OpenXT/xenclient-oe/pull/1101 for a comment on supporting the upgrade case.

templates/default/service-ndvm-hvm
```diff
@@ -23,7 +23,8 @@
     "1": "myself -> 0:5555",
     "2": "0:15555 -> myself:5555",
     "3": "myself -> 0:4878",
-    "4": "my-stubdom -> 0:5100"
+    "4": "my-stubdom -> 0:5100",
+    "5": "myself -> 0:5556"
```
jandryuk: Would it be better to change the qtdbd tools to just use the restricted port 5555?

xenmgr/Vm/Templates.hs
```diff
@@ -57,7 +57,7 @@ import Paths_xenmgr
 
 data ConfigTemplate = ConfigTemplate JSValue
 
-ndvmDefaultMode = "pv"
+ndvmDefaultMode = "hvm"
```
jandryuk: Is a hvm ndvm required for 64bit?  I'm not opposed, but why is this change being made?

templates/default/service-ndvm-hvm
crogers1: Yea we can do that, I'll make the change

xenmgr/Vm/Templates.hs
crogers1: Better to just use hvm until we can get pvh support because set of PRs makes everything 64-bit, including the NDVM.  If we don't want a x64 NDVM (or UIVM) we should have that discussion.

crogers1:
Will squash and cleanup the commits here.  I'll also note that aside from the hvm ndvm change, this PR is a cleanup PR like the one for idl.git.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6446

xenmgr/Vm/Templates.hs
crogers1: Actually as a follow up, I'm going to test how well pvh UIVM works on x64

jean-edouard: This change makes a lot of sense, especially for a 64bits guest.
However, I'd recommend to do it in a separate PR, for a couple of reasons:
- Even if it's a one-liner, it's a big change, I think it deserves its own PR, build and tests.
- I have noticed that NDVM is slower to start as an HVM, especially on first boot after install, to the point where the user has to wait ~30 seconds before creating a first VM, otherwise the network options may not be available.

jandryuk: @crogers1 My question is really, does a 64bit PV NDVM work?

@jean-edouard yes, it is slower.  I didn't think was *that* slow, but looking at a local machine, yes it is.  NDVM boots first though, so while everything waits for it, networking is available when starting guest.  Have you seen something different?

jean-edouard: I'm really talking about creating a guest, not starting it.

crogers1: @jandryuk yes, in the testing we've done so far, a 64bit PV NDVM does work.
@jean-edouard I've noticed something similar but haven't tracked down the cause yet.  Since nothing was technically broken I've been focused on other issues.

crogers1:
This PR isn't necessary anymore.  The pv -> hvm NDVM change will follow in another PR after x64 is merged.  It's necessary for stable-9, but not for x64 itself.

PR #137, opened 2019-04-17
OXT-1556: fix iso-hotswap

mahantesh-ais:
OXT-1556

For cdrom devices keep the name as 'hdc' because this is
used in iso-hotswap code.
AHCI currently doesn't support read-only devices, hence cdrom devices
cannot be AHCI emulated.

Signed-off-by: Mahantesh Salimath <salimathm@ainfosec.com>


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6522

jean-edouard:
LGTM, merging soon.

PR #139, opened 2019-04-24
OXT-1538: Re-work partition mounting.

jandryuk:
This PR re-works mounting disk images to address the intermittent start failure of NDVM and other VMs.  A kernel change in 4.19 introduced an intermittent failure when using an offset with mount.  Under the hood, mount creates a loopdev when using an offset.

The old xenmgr code to mount a partition created a tapdev and then used mount -o offset= to mount a partition.  The new way, creates the tapdev, creates a loopdev with partitions, and then mounts the loopdev partition.  This side-steps the offset issue and lets the mounting succeed.

Using loopdev partition scanning is a little nicer in that we don't have to run fdisk and parse that output.  We just let the kernel do its thing and use the created device nodes.

PR #140, opened 2019-04-24
STABLE-9: OXT-1538: Re-work partition mounting

jandryuk:
This is the Stable-9 version of https://github.com/OpenXT/manager/pull/139

This PR re-works mounting disk images to address the intermittent start failure of NDVM and other VMs. A kernel change in 4.19 introduced an intermittent failure when using an offset with mount. Under the hood, mount creates a loopdev when using an offset.

The old xenmgr code to mount a partition created a tapdev and then used mount -o offset= to mount a partition. The new way, creates the tapdev, creates a loopdev with partitions, and then mounts the loopdev partition. This side-steps the offset issue and lets the mounting succeed.

Using loopdev partition scanning is a little nicer in that we don't have to run fdisk and parse that output. We just let the kernel do its thing and use the created device nodes.


eric-ch:
Building: [$6531](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6531)

jean-edouard:
Merging soon.

jean-edouard:
@jandryuk, please align with master PR.

jandryuk:
Updated.  Thanks for checking and the reminder.

jean-edouard:
Merging soon.

PR #141, opened 2019-04-25
xenmgr: Reconnect VIFs manually

jandryuk:
We have domains providing network backends that do not run
network-slave.  In those cases, when the VM restarts, network-daemon
repeatedly tries to call network-slave resulting in eventual failure.
network-daemon: domain's 12 sytem bus is unresponsive: connect: invalid argument (Transport endpoint is not connected), retrying..

If we don't specify a "network" in the nic config, xenmgr won't try to
reconnect frontends.  Change that to instead call Xl.setNicBackendDom
when network is empty and bypass talking to network daemon/slave.  That
way we'll reconnect and avoid syslog messages.

### DIFF
--- a/xenmgr/Vm/Dm.hs
+++ b/xenmgr/Vm/Dm.hs
@@ -153,7 +153,8 @@ moveBackend t frontdomid id backdomid = do
              case (nicNet /= "") of
                  True   -> do liftIO $ Xl.removeNic uuid id backdomid
                               vifConnect uuid id nicNet frontdomid backdomid 30
-                 False  -> return ()
+                 False  -> do liftIO $ Xl.setNicBackendDom uuid id backdomid
+                              return ()
       -- Try to hook up the vif to the backend, retrying for specified timeout in seconds
       -- Rpc calls are also wrapped in their own retry block in case dbus isn't ready in the ndvm
       vifConnect uuid id nicNet frontdomid backdomid timeout =


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6533

stacktrust:
Is this a bugfix or a new feature?

jandryuk:
It's (sort of) restoring functionality that worked in older releases.  xenvm re-attached VIFs unilaterally and didn't talk to network-daemon.  The xenmgr xl changes only re-attached VIFs with a "network" definition - plus it makes an RPC call to network-daemon/network-slave to check on the status.  This change re-attaches VIFs without a "network" definition while not talking to network-daemon in that case.

stacktrust:
Would you mind drafting/reviewing/revising a release note for this change, either here or in a JIRA ticket?  E.g.

"For custom NDVM images which do not contain the network-slave application, modify the VM configuration for connected guests as follows ..."

jandryuk:
For custom NDVMs and service VMs that do not include network-slave, `db-rm /vm/$uuid/config/nic/$N/network` for any guests using network-slave-less network backends.

When a NDVM or service vm is restarted, xenmgr will re-attched any Xen netfront drivers to the new network backends.  To confirm attachment, xenmgr will RPC through network-daemon to network-slave for any NICs with a defined "network" (`/vm/$uuid/config/nic/$N/network`).  If the NDVM/service VM does not run network-slave, `db-rm` the "network" key to avoid triggering RPC calls which will go unanswered.

jean-edouard:
Merging soon.

PR #142, opened 2019-04-26
xenmgr: Print the uuid during state checks

jandryuk:
"active vms" isn't informative.  Add the uuid to the output.

Signed-off-by: Jason Andryuk <jandryuk@gmail.com>

PR #143, opened 2019-04-26
Quiet rpc-proxy

jandryuk:
This PR changes lots of messages to debug priority which makes rpc-proxy less chatty in /var/log/messages.  They are at debug level, so rsyslog sees them and then chooses to ignore them with out default /etc/rsyslog.conf.

For a single message, it changes from
> rpc-proxy: incoming connection from domain 1, port 0x9e366f5e, fd is 10
> rpc-proxy: null byte received from incoming transport - setting up conversation
> rpc-proxy: authenticating client <v4v: 10> ArtefactSource {sourceDomainID = 1, sourceUuid = Just 00000000-0000-0000-0000-000000000002, sourceIsStubdom = False} rpc-proxy: auth: null byte sent to server
> rpc-proxy: auth done for <v4v: 10>
> rpc-proxy: starting forwarding <socket: 14> -> <v4v: 10>
> rpc-proxy: starting forwarding <v4v: 10> -> <socket: 14>
> rpc-proxy: assuming (00000000-0000-0000-0000-000000000002,domstore-read-access) = True
> rpc-proxy: assuming (00000000-0000-0000-0000-000000000002,domstore-write-access) = True
> rpc-proxy: assuming (00000000-0000-0000-0000-000000000002,usb-control) = False
> rpc-proxy: assuming (00000000-0000-0000-0000-000000000002,type) = "ndvm"
> rpc-proxy: <v4v: 10>: end of stream reached on <v4v: 10>
> rpc-proxy: stopping forwarding <v4v: 10> -> <socket: 14>
> rpc-proxy: <socket: 14>: end of stream reached on <socket: 14>
> rpc-proxy: stopping forwarding <socket: 14> -> <v4v: 10> 

to:
> rpc-proxy: incoming connection from domain 1, port 0x9e366f5e, fd is 10
> rpc-proxy: authenticating client <v4v: 10> ArtefactSource {sourceDomainID = 1, sourceUuid = Just 00000000-0000-0000-0000-000000000002, sourceIsStubdom = False}

### DIFF
--- a/rpc-proxy/Msg/DBus.hs
+++ b/rpc-proxy/Msg/DBus.hs
@@ -97,6 +97,8 @@ incomingData sock =
     ) `E.catch` err
   where
     err :: E.SomeException -> IO LazyBS.ByteString
-    err x = do warn $ printf "%s on %s" (show x) (show sock)
+    err x = do _err $ E.fromException x
                return LazyBS.empty
-
+    _err (Just (EOF s)) = debug $ printf "%s on %s" (show (EOF s)) (show sock)
+    _err (Just ex) = warn $ printf "%s on %s" (show ex) (show sock)
+    _err Nothing = return ()
--- a/rpc-proxy/Firewall.hs
+++ b/rpc-proxy/Firewall.hs
@@ -126,7 +126,7 @@ augmentRule (client, Just uuid, pmap) rule@Rule { match = match } = do
               Nothing -> do
                 -- query current property value via xenmgr access
                 v <- testVm (vmPath uuid) pname
-                info $ "assuming (" ++ show uuid ++ "," ++ pname ++ ") = " ++ show v
+                debug $ "assuming (" ++ show uuid ++ "," ++ pname ++ ") = " ++ show v
                 return ( Map.insert (uuid,pname) (cons v) pm
                        , cons v )
       
--- a/rpc-proxy/Msg/Json.hs
+++ b/rpc-proxy/Msg/Json.hs
@@ -148,8 +148,11 @@ incomingData sock =
     ) `E.catch` err
   where
     err :: E.SomeException -> IO BL.ByteString
-    err x = do warn $ printf "%s on %s" (show x) (show sock)
+    err x = do _err $ E.fromException x
                return BL.empty
+    _err (Just (EOF s)) = do debug $ printf "%s on %s" (show (EOF s)) (show sock)
+    _err (Just ex) = do warn $ printf "%s on %s" (show ex) (show sock)
+    _err Nothing = do return ()
 
 jmsgToJson :: JMsg -> JSValue
 jmsgToJson (JMsgReq m) = jreqToJson m
--- a/rpc-proxy/RpcProxy.hs
+++ b/rpc-proxy/RpcProxy.hs
@@ -181,14 +181,14 @@ conversation dispatcher counter settings rulesCache (client_ch,client_source) an
         null_buf <- recv client_ch 1
         when ( BS.head null_buf /= 0 )
           $ E.throw (AuthError client_ch "expected null byte")
-        info "null byte received from incoming transport - setting up conversation"
+        debug "null byte received from incoming transport - setting up conversation"
         (forward_ch, forward_domid) <-
           makeOutgoingTransport (outgoingChannel settings)
         debug $ "outgoing transport: " ++ show forward_ch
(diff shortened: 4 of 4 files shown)


jandryuk:
Thanks, @crogers1 .  Are you okay with this and https://github.com/OpenXT/manager/pull/142 for stable-9?

crogers1:
Yep!

jandryuk:
Thanks, @crogers1

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6533

rpc-proxy/Msg/DBus.hs
```diff
@@ -97,6 +97,10 @@ incomingData sock =
...
-    err x = do warn $ printf "%s on %s" (show x) (show sock)
-               return LazyBS.empty
-
+    err x = do
+      case E.fromException x of
+       Just (EOF s) -> do debug $ printf "%s on %s" (show x) (show sock)
+                          return LazyBS.empty
+       Just ex -> do warn $ printf "%s on %s" (show x) (show sock)
+                     return LazyBS.empty
+       Nothing -> do return LazyBS.empty
```
jean-edouard: Couldn't this be pattern matched? like:
```
err Just (EOF s) = do debug [...]
err Just ex = do warn [...]
err Nothing = do [...]
```
Also, `Just ex` should be replaceable by "Just _", since the value doesn't seem to be used.

crogers1: Can't be pattern matched like you have above because you need to call E.fromException.  ```Just ex``` can certainly be replaced by _.

jean-edouard: Ack tx, I'll never understand haskell :)

jandryuk: These two options compile.   I haven't tested through and don't know if changing `show x` to `show s` or `show ex` change the output.  Are either preferred, or should I just make the `Just _` change?

```
    err :: E.SomeException -> IO LazyBS.ByteString
    err x = do _err $ E.fromException x
                return LazyBS.empty
    _err (Just (EOF s)) = debug $ printf "%s on %s" (show s) (show sock)
    _err (Just ex) = warn $ printf "%s on %s" (show ex) (show sock)
    _err Nothing = return ()
```
or
```
    err :: E.SomeException -> IO BL.ByteString 
    err x = do
      case E.fromException x of
        Just (EOF s) -> do debug $ printf "%s on %s" (show x) (show sock)
        Just ex -> do warn $ printf "%s on %s" (show x) (show sock)
        Nothing -> do return ()
      return BL.empty
```

jean-edouard: They both look better IMO.
I like the first one because it's a bit more "functional", and the second one because it's cleaner.
Sorry for nit picking on this, honestly the original one is fine too if you don't have time to test another option.

jandryuk: I kicked off a build of the first option and will test it tomorrow.

crogers1: Idk what kind of output you'll get for the (show ex), I guess it'll try to print the Error it was converted to instead of the Exception it was originally, both classes clearly derive Show so it should work.

jandryuk: On the new code, I had to do:
`_err (Just (EOF s)) = do debug $ printf "%s on %s" (show (EOF s)) (show sock)` 
with just `(show s)` it was only printing the socket and skipped "end of stream received".
> <v4v: 9> on <v4v: 9>

instead of
> <v4v: 14>: end of stream reached on <v4v: 14>

Yeah, I'm not sure what:
`_err (Just ex) = do warn $ printf "%s on %s" (show ex) (show sock)`
will print since it prints the value after fromException.  Channel.hs defines `Show BounceErr` that formats the Exceptions, so we'd probably lose that...

jandryuk: I dropped the `EOF` pattern and made everything go through `_err (Just ex) = warn $ printf "%s on %s" (show ex) (show sock)`, and the output was unchanged.  We should be good here.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6539

jean-edouard:
LGTM, merging soon.

PR #146, opened 2019-05-06
OXT-1596: [xenmgr] Error if VMs disk paths are bad

crogers1:
When a the phys-path to a VM's disk is bad, the VM can be stuck
  for a long time in an invalid state without a visual indication
  something is wrong. This will also catch the case where a phys-path
  is empty or the actual disk was deleted or moved. Xenmgr will throw
  an error which can be seen in the UI and the VM state will be
  reset to a sane state of Shutdown.

  OXT-1596


jean-edouard:
Merging soon.

PR #148, opened 2019-05-06
STABLE-9: OXT-1596: [xenmgr] Error if VMs disk paths are bad

crogers1:
When a the phys-path to a VM's disk is bad, the VM can be stuck
  for a long time in an invalid state without a visual indication
  something is wrong. This will also catch the case where a phys-path
  is empty or the actual disk was deleted or moved. Xenmgr will throw
  an error which can be seen in the UI and the VM state will be
  reset to a sane state of Shutdown.

  OXT-1596

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -151,7 +151,7 @@ import System.IO
 import System.Timeout
 import Data.Time
 import Directory
-import System.Directory (createDirectoryIfMissing)
+import System.Directory (createDirectoryIfMissing, doesFileExist)
 import qualified Data.Foldable
 
 import Tools.Log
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6546

jean-edouard:
Works great, one minor comment though, the message says "bad path" twice...
Looks like: "Bad Physical Path for disk 0 bad path /storage/isos/nulll.iso"
How about something like: "Bad Physical Path for disk 0 (/storage/isos/nulll.iso)"
Thanks

crogers1:
Sure I can update that

jean-edouard:
Merging soon.

PR #149, opened 2019-05-09
STABLE-9: xenmgr: Reconnect VIFs manually

jandryuk:
Stable-9 version of https://github.com/OpenXT/manager/pull/141

We have domains providing network backends that do not run
network-slave.  In those cases, when the VM restarts, network-daemon
repeatedly tries to call network-slave resulting in eventual failure.
network-daemon: domain's 12 sytem bus is unresponsive: connect: invalid argument (Transport endpoint is not connected), retrying..

If we don't specify a "network" in the nic config, xenmgr won't try to
reconnect frontends.  Change that to instead call Xl.setNicBackendDom
when network is empty and bypass talking to network daemon/slave.  That
way we'll reconnect and avoid syslog message.

(cherry picked from commit 7779984634616e93d486ceb2674ef64cd74206d1)


jean-edouard:
Merging soon.

PR #150, opened 2019-05-09
xenmgr: Add xl trigger power to HVM shutdown

jandryuk:
The QEMU ACPI hvm-shutdown and hvm-power-button xenstore entries don't
always work.  On the QEMU side, it looks like the IRQ is raised and the
ACPI_SLP_BIT is cleared, but the guest doesn't shutdown.  `xl shutdown
-F -w` just hangs.

Manually running `xl trigger DomU power` after that seems to trigger the
shutdown, so just have xenmgr make that invocation.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6555

jean-edouard:
LGTM, merging soon.

PR #152, opened 2019-05-15
xenmgr: Fix Power control affects VM and Host

jandryuk:
When a VM was set for "Power control affects VM and host", I was seeing
a guest reboot end up shutting down the host.  When "shutdown vm" is
called, the shutdown reason reason was observed to be Halt or Reboot - I
haven't seen Restarting.  Check for Halt and shutdown the host in that
case.  That could miss the AcpiPowerOff reason, but that has not been
seen in practice.

@crogers1 you set this to the original 'reason /= Restarting'.  Is there something I am missing?

'data ShutdownReason = CreationFailure | Reboot | Hibernate | Halt | Restarting | AcpiPoweroff'

To be clear, in older versions a guest initiated reboot did not affect the host power state - only the guest rebooted.

### DIFF
--- a/xenmgr/Vm/React.hs
+++ b/xenmgr/Vm/React.hs
@@ -208,7 +208,8 @@ powerlinkR xm get_shr =
                             liftIO (Xl.resumeFromSleep (vm_uuid vm))
                             return ()
       shutdown vm      = do reason <- runVm vm get_shr
-                            when ( reason /= Restarting ) $
+                            debug $ "Power Link: shutdown hook reason " ++ show reason
+                            when ( reason == Halt ) $
                               info "Power Link: shutdown" >> hostShutdown
 
 logStatesR = mkReact f where


crogers1:
There's an extra state called "Rebooting" defined in xenmgr that wasn't originally hooked up prior to the libxl merge for some reason.  The XenMgr state machine in general is a rough piece of work, but I felt that using the "Rebooting" state would provide some extra information about guests and could be used for improving existing features.  The immediate benefit was the visual update in the UIVM that shows a vm "Restarting" while it's turning off instead of just On -> Shutdown.

I believe I made the reason /= Restarting change because when the powerlink react code runs the shutdown reason isn't always AcpiPoweroff with our libxl patches.  It can be Halt, and the powerlink wasn't triggering.  I must have observed the shutdown reason to be Restarting at some point during testing and clearly didn't want to disallow that.  I didn't think it was possible to fall into the powerlink shutdown handler with a Shutdown Reason of "Reboot" due to the (VmStateChange Shutdown) requirement.  

But I think this commit broke it: https://github.com/OpenXT/manager/commit/26695ea8e64966991affef1d996bd016e21e4615
by translating a "rebooted" state event to "shutdown" to get the right cleanup handlers running.

jandryuk:
Thanks, @crogers1 

What do you think about doing `reason == Halt || reason == AcpiPoweroff`?

Or how about removing AcpiPoweroff?  The two uses in the code are the definition and ```
`mappend` whenE (VmAcpiStateChange 5)     ( set_shr AcpiPoweroff )```  Switch the `set_shr` to `Halt` and drop `AcpiPoweroff`.  Doing that, we'd start running the `Halt` code, but shouldn't it be called for `AcpiPoweroff`?

Also, I've not seen `AcpiPoweroff` in my (limited) testing.

crogers1:
I'd be ok with the ```||``` case.  As far as AcpiPoweroff is concerned, that does look to be the only use inside xenmgr, also the only code that reacts to ```(VmAcpiStateChange 5)```, though other daemons may listen for that acpi_state message.  It's _probably_ ok to replace AcpiPoweroff with Halt in React.hs, but off the top of my head I don't know if it will affect anything else.

eric-ch:
Build [$6559](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6559)

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6573

jandryuk:
I'm going to leave this PR as-is and not change for `AcpiPoweroff`.  Questions about that only came up while looking at the code, and I haven't actually seen a problem.  To avoid unintended consequences, I'll just leave it be.

crogers1:
@jandryuk I'm ok with that.  We can open an improvement ticket to try removing it later.

jean-edouard:
I'm sorry but I'm very unclear on what this PR fixes...
Is it that with "Power control affects VM and host" guest reboot doesn't affect the host?
That doesn't seem to be the case with or without this PR.
Please give me a quick expected vs observed behavior.
Thanks

jandryuk:
@jean-edouard Windows 7 Guest without tools & set to "Power control affects VM and host"
In the guest, trigger a reboot.
Observed behavior: VM shuts down and then host shuts down.
Expected behavior: Only the VM reboots

jean-edouard:
Thank you @jandryuk, this fix works as expected, merging soon.

PR #153, opened 2019-05-15
xenmgr: Use /dev/hvc0 for measurement failure messages

jandryuk:
/dev/xvc0 doesn't exist on para-virt Xen kernels.  Use /dev/hvc0 to
actually print out the message.

### DIFF
--- a/xenmgr/Vm/React.hs
+++ b/xenmgr/Vm/React.hs
@@ -159,7 +159,7 @@ measuresR xm_context = mkReact f where
   f (VmMeasurementFailure fpath expected actual)
     = do action <- liftRpc getMeasureFailAction
          let msg = printf "measuring of file %s FAILED. Expected checksum: %x actual %x. Performing %s" fpath expected actual (show action)
-         liftIO (warn msg >> writeFile "/dev/xvc0" (msg++"\n"))
+         liftIO (warn msg >> writeFile "/dev/hvc0" (msg++"\n"))
          runXM xm_context (executePmAction action)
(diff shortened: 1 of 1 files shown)


eric-ch:
Build [$6559](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6559)

xenmgr/Vm/React.hs
```diff
@@ -159,7 +159,7 @@ measuresR xm_context = mkReact f where
   f (VmMeasurementFailure fpath expected actual)
     = do action <- liftRpc getMeasureFailAction
          let msg = printf "measuring of file %s FAILED. Expected checksum: %x actual %x. Performing %s" fpath expected actual (show action)
-         liftIO (warn msg >> writeFile "/dev/xvc0" (msg++"\n"))
+         liftIO (warn msg >> writeFile "/dev/hvc0" (msg++"\n"))
```
eric-ch: No complaint about the change, but why is this logged to the serial output directly? Shouldn't this be sent to the system logs?

jandryuk: It is logged to syslog through `warn msg`.  I guess it's also written to hvc0 so there is some notification to the user who wouldn't be able to access the syslog messages.

eric-ch:
Will merge soon.

eric-ch:
@jandryuk Can you send a PR against stable-9 as well please?

jandryuk:
stable-9 PR posted.

PR #103, opened 2017-09-12
xenmgr: use readProcessOrDie to avoid fd leaks

jean-edouard:
OXT-1193

Signed-off-by: Jed <lejosnej@ainfosec.com>

PR #155, opened 2019-05-29
xenmgr: avoid leaking FDs to run scripts

jandryuk:
runInteractiveProcess doesn't touch FDs, so it leaks them out into the
wild.  This is seen by AVCs for /etc/xenclient.conf and a socket or pipe
when a shell script runs.

Switch to createProcess so we can set close_fds = True and avoid the
leakage.

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -301,7 +301,10 @@ runEventScript failmode vm get_hooks args =
          return ()
     exec path =
       -- stderr redirected to warn, stdout to info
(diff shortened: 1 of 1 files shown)


jean-edouard:
Sorry I missed this one in https://github.com/OpenXT/manager/pull/103 :|

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6579

jean-edouard:
LGTM, merging soon.

PR #156, opened 2019-05-30
S9: xenmgr: Fix Power control affects VM and Host

jandryuk:
This is the Stable-9 version of https://github.com/OpenXT/manager/pull/152

When a VM was set for "Power control affects VM and host", I was seeing
a guest reboot end up shutting down the host.  When "shutdown vm" is
called, the shutdown reason reason was observed to be Halt or Reboot - I
haven't seen Restarting.  Check for Halt and shutdown the host in that
case.  That could miss the AcpiPowerOff reason, but that has not been
seen in practice.

(cherry picked from commit 20c7e134857ec71a68009c2575f2f8cbe67f2847)


jean-edouard:
Merging soon.

PR #157, opened 2019-05-31
S9: xenmgr: Add xl trigger power to HVM shutdown

jandryuk:
This is the Stable-9 version of https://github.com/OpenXT/manager/pull/150

The QEMU ACPI hvm-shutdown and hvm-power-button xenstore entries don't
always work.  On the QEMU side, it looks like the IRQ is raised and the
ACPI_SLP_BIT is cleared, but the guest doesn't shutdown.  `xl shutdown
-F -w` just hangs.

Manually running `xl trigger DomU power` after that seems to trigger the
shutdown, so just have xenmgr make that invocation.

(cherry picked from commit 1cfb6aa8e63e45f09b36690bca09128bebad465a)


jean-edouard:
Merging soon.

PR #158, opened 2019-05-31
S9: xenmgr: avoid leaking FDs to run scripts

jandryuk:
This is the stable-9 version of https://github.com/OpenXT/manager/pull/155

runInteractiveProcess doesn't touch FDs, so it leaks them out into the
wild.  This is seen by AVCs for /etc/xenclient.conf and a socket or pipe
when a shell script runs.

Switch to createProcess so we can set close_fds = True and avoid the
leakage.

(cherry picked from commit 1a04967c637c65d996ab7d80dbeae304a5872bc0)


jean-edouard:
Merging soon.

PR #159, opened 2019-05-31
xenmgr: Close some FD leaks

jandryuk:
xl create lives for the lifetime of the domain, so we don't want it to
have our FDs.  The other commands are one-offs, but they don't need to
inherit FDs.

These commands remain in the same SELinux domain (type), so they don't
cause avcs.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6584

jean-edouard:
LGTM, merging soon.

PR #160, opened 2019-05-31
S9: xenmgr: Close some FD leaks

jandryuk:
This is the stable-9 version of https://github.com/OpenXT/manager/pull/159

xl create lives for the lifetime of the domain, so we don't want it to
have our FDs.  The other commands are one-offs, but they don't need to
inherit FDs.

These commands remain in the same SELinux domain (type), so they don't
cause avcs.

(cherry picked from commit b09ad99476cb232f026c95b2be738ac574ccb728)


jean-edouard:
Merging soon.

PR #165, opened 2019-06-11
OXT-1582: [linux] Use ide instead of AHCI

crogers1:
For Linux guests, with AHCI emulation there will be 2 disks
  shown to the guest; the AHCI emulated and the PV blkfront which
  is compiled into most modern Linux distros. Have Linux guests
  use IDE emulation so the Xen/Qemu unplug path can remove the
  emulated disk to avoid confusing the user. AHCI vs IDE doesn't
  matter because the PV driver will be used anyway.

  OXT-1582

### DIFF
--- a/templates/default/new-vm-linux
+++ b/templates/default/new-vm-linux
@@ -26,7 +26,7 @@
     "display": "none",
     "boot": "cd",
     "flask-label": "system_u:system_r:hvm_guest_t",
-    "hdtype": "ahci",
+    "hdtype": "ide",
     "disk": {
       "0": {
         "path": "\/storage\/isos\/xc-tools.iso",


crogers1:
The code path doesn't exist from the Linux kernel (in guest) to QEMU, like it does for IDE.  If you send a QMP command via xl to pop the AHCI device, it _will_ remove it, but timing becomes an issue.  For example, it won't remove it if the guest is waiting to boot in seabios, but will if you send the command once you're into the guest installer.  But telling the toolstack when the guest is precisely in that state so it can send the command is non-trivial.

jean-edouard:
From what I understand, unplugging IDE drives right after the guest loads blkfront is an OpenXT-specific hack that is not desired upstream.
I assume this PR is just to make Linux guests foolproof, in case some crazy admin would decide to mount both the emulated and pv disk in read-write at the same time.

eric-ch:
Build [$6618](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6618)

jean-edouard:
LGTM, merging soon.

PR #167, opened 2019-06-11
[argo] Replace v4v with argo

tsirakisn:
- Basic find and replace. All references to v4v in
  function names, variables, comments, etc. were
  changed to their argo equivalents.
- Update invalidDomId from 0x7fff to 0x7ff4 (defined
  in upstream Xen).

OXT-1464

Signed-off-by: Nicholas Tsirakis <tsirakisn@ainfosec.com>


dozylynx:
Looks good.

crogers1:
> I don't think it makes sense to modify the old migration scripts.
> Instead, a new migration script should be written to `s/v4v-firewall-rules/argo-firewall-rules/`.
> In its current state, I think this PR breaks upgrade.

Yea Niko and I talked offline. Your assessment is correct @jean-edouard.  There should be an update to this PR coming soon.

dozylynx:
> I don't think it makes sense to modify the old migration scripts.
> Instead, a new migration script should be written to `s/v4v-firewall-rules/argo-firewall-rules/`.
> In its current state, I think this PR breaks upgrade.

Good point, @jean-edouard

tsirakisn:
Addressed @jean-edouard 's comment.

jean-edouard:
Don't you want to also rename `/config/v4v` in M_40?

tsirakisn:
@jean-edouard ugh long day...

tsirakisn:
Added /config/v4v to migration

PR #170, opened 2019-07-23
OXT-1647: templates: Make UIVM default to PVH.

eric-ch:
Change `virt-type` from `pv` to `pvh`.

### DIFF
--- a/templates/default/service-uivm
+++ b/templates/default/service-uivm
@@ -63,7 +63,7 @@
     "debug": "true",
     "pae": "true",
     "acpi": "true",
-    "virt-type": "pv",
+    "virt-type": "pvh",
     "apic": "true",
     "nx": "true",
     "vkbd" : "true",


eric-ch:
### Required PRs:

- https://github.com/OpenXT/xenclient-oe/pull/1219
- https://github.com/OpenXT/input/pull/34
- https://github.com/OpenXT/surfman/pull/28

### Optional:
- https://github.com/OpenXT/manager/pull/170 (Default UIVM to pvh).

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6668

jean-edouard:
LGTM but did you test upgrade?

eric-ch:
@jean-edouard embarrassingly not... Will do so. If someone else does, my main concern would be that the updated installation did not switch to PVH. The UIVM kernel still has the full PV support, so it will boot fine as PV.

jean-edouard:
Just tested it, works fine, merging soon.

PR #172, opened 2019-08-14
templates: Switch xencons to console

jandryuk:
xencons is the old option name which has no affect on modern linux.  Use
console to specify the console.

This will restore console output in /var/log/messages from uivm and
syncvm.


jean-edouard:
Build here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6690

PR #173, opened 2019-08-21
OXT-1684: upgrade-db: Update db-version to apply Migration_40

jandryuk:
Upgrades are not handling the v4v->argo conversion since Migration_40 is
not being run.  We need to set latestVersion to 41 to run the
migrations.

NDVM & UIVM don't have a problem since their configs are injected from
on-disk copies already updated for argo at boot time.  It's only other
existing VMs with v4v rules that need updating.

Signed-off-by: Jason Andryuk <jandryuk@gmail.com>

PR #174, opened 2019-08-21
S9: OXT-1684: upgrade-db: Update db-version to apply Migration_40

jandryuk:
Stable-9 version of https://github.com/OpenXT/manager/pull/173

Upgrades are not handling the v4v->argo conversion since Migration_40 is
not being run.  We need to set latestVersion to 41 to run the
migrations.

NDVM & UIVM don't have a problem since their configs are injected from
on-disk copies already updated for argo at boot time.  It's only other
existing VMs with v4v rules that need updating.

(cherry picked from commit 551206c0da0ba6adc03356005cecfc1a2c7442ba)

### DIFF
--- a/upgrade-db/Upgrade.hs
+++ b/upgrade-db/Upgrade.hs
@@ -40,7 +40,7 @@ import qualified Data.Text as T
 
 -- MODIFY THIS WHEN FORMAT CHANGES
 latestVersion :: Int
-latestVersion = 40
+latestVersion = 41
 ----------------------------------
 
 dbdRunning :: IO Bool


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6713

stacktrust:
Please add ticket number with target version in JIRA metadata.   How do we audit for other potential gaps in 8-9 upgrade?  What other features were impacted by the missing db upgrade?

eric-ch:
[OXT-1684](https://openxt.atlassian.net/browse/OXT-1684)

eric-ch:
Will merge soon.

PR #175, opened 2019-12-18
xenmgr: Use device-model directory for shutdown

jandryuk:
xenclient-oe commit 6dde8833f7de "xen: Remove duplicate xenstore writes"
changed a domain's control xenstore directory to read-only.  This broke
stubdom qemu's ability to write to "control/hvm-powerbutton-enable".
Without that node, xenmgr wouldn't use the acpi shutdown paths.

Switch to writing hvm-powerbutton-enable and hvm-shutdown in the
device-model directory of qemu-dm's domain.  That is either the stubdom
or dom0.  Logically it makes sense to put this device-model information
underneath the device-model instead of the guest.  It also side steps
the permission issue.

Goes along with https://github.com/OpenXT/xenclient-oe/pull/1304

### DIFF
--- a/xenmgr/XenMgr/Connect/Xl.hs
+++ b/xenmgr/XenMgr/Connect/Xl.hs
@@ -179,12 +179,14 @@ shutdown :: Uuid -> IO ()
 shutdown uuid =
     do
       domid <- getDomainId uuid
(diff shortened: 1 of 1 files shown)


crogers1:
Syntax looks good, I think storing the hvm-* keys in the device-model path is definitely better.

eric-ch:
Will merge soon.

PR #176, opened 2020-02-25
OXT-1686 : create-ndvm: enable HAP/EPT/SLAT

stacktrust:
For increased security, enable HAP (hardware-assisted paging) SLAT
(Second Level Address Translation) instead of Xen shadow page tables.
Also known as:

  - Intel EPT (Extended Page Tables)
  - AMD NPT (Nested Page Tables)

OXT-1686

Signed-off-by: Rich Persaud <rich.persaud@baesystems.com>

Requires https://github.com/OpenXT/xenclient-oe/pull/1339

### DIFF
--- a/templates/default/new-vm-ndvm-hvm
+++ b/templates/default/new-vm-ndvm-hvm
@@ -40,6 +40,7 @@
     "acpi": "true",
     "virt-type": "hvm",
     "apic": "true",
+    "hap": "true",
     "nx": "true",
     "argo": "true",
     "memory": "192",


jandryuk:
Xen enables `hap` by default if available.  Will this prevent NDVM from starting if the platform doesn't support hap?

I don't think there is actually any requirement on OpenXT/xenclient-oe#1339

eric-ch:
Build [$10071](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10071)

eric-ch:
As @jandryuk said, hap is enabled by default when available[[1]](https://xenbits.xen.org/docs/unstable/man/xl.cfg.5.html), do you have any suspicion this is not the case?
The default guest templates do force `"hap": "true"`

stacktrust:
Even though upstream enables HAP by default, enabling HAP in OpenXT's NDVM config has the effect of documenting OpenXT's security requirement for hardware-assisted paging.  Upstream Xen's security support for NDVMs is contingent on HAP.  Ideally the HAP config property would be visible to local UIVMs and remote configuration management systems.

Since we already enable HAP in OpenXT guest VM templates, there's no downside to enabling HAP in NDVM templates.  If we need to run NDVMs without HAP, e.g. testing under Qemu, that can be enabled in a 'dev/debug' build of OpenXT.

Separately, we should consider disabling software shadow page tables in Xen.  Systems without HAP are not security supported by OpenXT or Xen with PCI passthrough.

jandryuk:
You didn't answer the question: does setting hap=1 prevent booting when hap is not available?

Where is it stated that HAP is an OpenXT security requirement?

How is upstream security support contingent on HAP?

jandryuk:
As @eric-ch points out, `"hap": "true"` is set for the guest VM templates, so bringing NDVM into alignment is fine.

The commit message is inaccurate since HAP is already used.  It should be updated to state that HAP is now required for the NDVM.

For the record, it looks like Xen will not start a VM with `hap=1` when HAP is not available.

PR #177, opened 2020-02-25
OXT-1752 : templates: disable Argo for user VMs

stacktrust:
Since Argo is primarily used by service VMs, it is disabled by XSM
for guest VMs.  When enabled in the VM config file, Argo causes an
unknown device to appear in Windows Device Manager.  Disable argo in
VM config, to avoid alarming guest monitoring agents.

OXT-1752

Signed-off-by: Rich Persaud <rich.persaud@baesystems.com>

### DIFF
--- a/templates/default/new-vm
+++ b/templates/default/new-vm
@@ -20,7 +20,7 @@
     "viridian": "true",
     "hap": "true",
     "nx": "true",
-    "argo": "true",
+    "argo": "false",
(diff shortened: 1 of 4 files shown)


eric-ch:
Build [$10071](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10071)

eric-ch:
Will merge soon.

PR #178, opened 2020-05-04
S9: OXT-1787: ocamlfind: use across the board for libraries

eric-ch:
ocamlfind is a frontend with the package manager.
dbd uses it to find the necessary options to pass to ocamlc/ocamlopt in
order to use the packages managed by findlib (ocamlfind
<ocamlc|ocamlopt> -package).  This is done by overriding the
OCAMLC/OCAMLOPT variables that are already used by ocaml.bbclass to pass
the cross environment options.

In order to avoid re-writing more build-system bits, use ocamlfind for
every library to invoke the byte-code and native compiler and pass
additional package (-linkpkg/-package) options through OCAMLCOPT and
OCAMLOPTOPT. This way OCAMLC and OCAMLOPT can define the specific cross
environment and ocamlfind can provide its services to include packages
through findlib.

### DIFF
--- a/dbd/Makefile
+++ b/dbd/Makefile
@@ -19,11 +19,8 @@
 include common.make
 
 PACKAGES = stdext,uuid,log,bigarray,camomile,json,jsonrpc,threads,unix,str,dbus,tscommon,xenmmap,xenbus,xenstore
-OCAMLOPT = ocamlfind ocamlopt -linkpkg -package $(PACKAGES)
-OCAMLC   = ocamlfind ocamlc -linkpkg -package $(PACKAGES)
-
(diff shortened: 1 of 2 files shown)


eric-ch:
Build [$7078](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7078)

eric-ch:
Build [$7084](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7084)

DBd builds and runs without problem on that build.
Now builds with `-j` provided by OE since the Makefile generates the correct dependencies (through `ocamldep`).

jandryuk:
Will merge soon.

PR #179, opened 2020-05-18
OXT-1787: dbd: Fix builds with findlib & parallelism

eric-ch:
- ocamlfind: use across the board for libraries
- Fix makefile dependency generation.


jandryuk:
Will merge soon.

PR #180, opened 2020-12-15
vglass: deprecate surfman logics and wire up vglass

eric-ch:
This changeset has 3 steps:
- Remove Surfman's integration and rules hardwired in the toolstack;
- Add vglass RPC and required logic to enable the vglass graphic stack;
- Enable vglass as the default graphic stack.

vglass support is enabled through the `vglass-enabled` configuration property, which is enabled by default for new guests and UIVM.

NOTE: PV and PVH guests that rely on openxtfb or glassdrm (e.g, UIVM) may experience a 30s boot time delay as xenbus will wait for the PV framebuffer/input devices to be connected to their backends. This is why `xenbus.wait_nonessentials` is provided and introduced by a Linux path in xenclient-oe.

NOTE: At the moment, emulated pointing devices do not work if PV tools are installed in Windows guests, this is why the configuration property `vkbd` is turned on by default to let `glass` handle a PV input instead.

### DIFF
--- a/apptool/samples/uivm.ovf
+++ b/apptool/samples/uivm.ovf
@@ -49,9 +49,6 @@
 	    <xci:RpcRule>allow destination com.citrix.xenclient.input</xci:RpcRule>
 	    <xci:RpcRule>allow destination com.citrix.xenclient.usbdaemon</xci:RpcRule>
 	    <xci:RpcRule>allow destination com.citrix.xenclient.updatemgr</xci:RpcRule>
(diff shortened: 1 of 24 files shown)


jandryuk (approved):
Generally looks good.

eric-ch:
### v2
- Remove `vfb` flag handling.

eric-ch:
#### V3
- Add Dbus calls to query/filter gpu devices

jandryuk (approved):

jandryuk:
Merging soon

PR #182, opened 2021-03-23
xenmgr: Avoid generating invalid configs

jandryuk:
It would be nice to have ndvm be able to toggle between hvm and pv by
just flipping the virt-type, but that isn't possible today.  xenmgr
needs to be more selective to only write appropriate options into the
xl.cfg for the corresponding virt-type.  This commit attempts to do so.

Centralize and expand the virt-type specific options in virtSpecs.

HVMs in OpenXT can't be direct booted, so don't provide kernel or
cmdline in that case.  xl throws an error when they are in xl.cfg for an
HVM.

"boot" is only relevant for HVM, but PV ignores it.

Note, we replace the use of "extra" with "cmdline" since that is
preferred and matches the option in.

Note 2: Disks should be specified by hda, and they will be reverted to
xvda for PV domains.

### DIFF
--- a/xenmgr/Vm/Config.hs
+++ b/xenmgr/Vm/Config.hs
@@ -559,7 +559,7 @@ stringifyXlConfig (XlConfig params) = unlines params
 getXlConfig :: VmConfig -> Rpc XlConfig
 getXlConfig cfg =
     fmap (XlConfig . concat) . mapM (force <=< future) $
-    [prelude, diskSpecs cfg, nicSpecs cfg, pciSpecs cfg
+    [prelude, virtSpecs cfg, diskSpecs cfg, nicSpecs cfg, pciSpecs cfg
     , miscSpecs cfg]
   where
     uuid = vmcfgUuid cfg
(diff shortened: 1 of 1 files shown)


jandryuk:
I changed the commit message to state the HVM direct boot situation more clearly:

While Xen supports direct booting HVMs with qemu placing the kernel in
VM memory, OpenXT doesn't have support for direct booting with Linux
stubdoms since we don't have a way to get the kernel to QEMU in the
stubdom to place into the VM.  Don't provide kernel or cmdline for HVMs
- xl throws an error when they are in xl.cfg for an HVM.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/9): [$master-custom-9](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/9/)

eric-ch (approved):

crogers1 (approved):
Looks good. Definitely gives finer grained control over these options.

eric-ch:
Will merge soon.

PR #184, opened 2021-06-15
RFC: rpc-proxy, xenmgr, updatemgr: Use rpc-autogen bindings module

eric-ch:
dbus-gen generates a set of bindings for the currently supported API. Use the built module instead of relying on generating the bindings from the IDL in the source tree.

### DIFF
--- a/rpc-proxy/rpc-proxy.cabal
+++ b/rpc-proxy/rpc-proxy.cabal
@@ -38,7 +38,8 @@ Executable rpc-proxy
     transformers-base,
     xchwebsocket,
     xchxenstore,
-    xchdb
+    xchdb,
+    rpc-autogen
   Main-Is: Main.hs
   if flag(stubdom)
     GHC-Options: -O2 -fwarn-incomplete-patterns -static -optl-static -optl-pthread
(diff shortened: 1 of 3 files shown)


rpc-proxy/rpc-proxy.cabal
```diff
@@ -38,7 +38,8 @@ Executable rpc-proxy
     transformers-base,
     xchwebsocket,
     xchxenstore,
-    xchdb
+    xchdb,
+    rpc-autogen
```
jandryuk, line 42: I'm surprised you had to add rpc-autogen.  xchdb switched to rpc-autogen in its own commit, so I would think that is sufficient.  Does rpc-proxy need some other idl binding that it was receiving transitively?

I don't know cabal, so I may be totally wrong.

eric-ch: rpcgen does have its dbus stub generated from an [xml](https://github.com/OpenXT/idl/blob/master/interfaces/rpc_proxy.xml#L6), which is why I put the explicit dependency, but it could get the object from xchdb. I can check that.

jandryuk:
Oh, I didn't realize rpc-proxy implemented an idl. Your change is good in
that case.

PR #191, opened 2021-11-02
xenmgr: Demote "active vm" messages to debug

jandryuk:
These can get excessive filling /var/log/messages.  It's unclear exactly
why, but make them debug so we don't see them.


crogers1 (approved):
Also LGTM

PR #192, opened 2022-04-27
Remove monitorAcpi thread

jandryuk:
xenmgr has a thread per-VM that polls the ACPI state once per second.  It actually forks two processes, xl uuid-to-domid and xl acpi-state.

Removing the thread noticably reduces xenmgr's CPU usage in top.  xenmgr can basically idle for 30 seconds between polling the disk usage.

All this ACPI stuff seems questionable.  I think it is mainly to support the run-on-acpi-state-change feature.  run-on-state-change seems better since the states are better handled by xenmgr.  I'd prefer to delete this ACPI state stuff if possible.

Also, VM S3 seems broken.  Windows 10 can't S3 as an OpenXT VM because the graphics subsystem prevents it.  However, I haven't checked if the vglass drivers allow it.  Linux detaches its PV disks on S3 resume, but the backend doesn't reconnect so it is basically dead.

S4 doesn't get reported out from Windows 10.  xenmgr just sees S5 poweroff.

I added code for VmStateChange Suspended, but that doesn't trigger.  If you S3 a Linux VM, it stays as "running".

Maybe these three commits should all be squashed together?

### DIFF
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -775,20 +775,6 @@ setupAcpiNode uuid =
      case stubdom of
          Just stubdomid -> liftIO $ xsChmod (xsp domid ++ "/acpi-state") ("b" ++ show stubdomid)
          Nothing        -> liftIO $ xsChmod (xsp domid ++ "/acpi-state") ("b" ++ show domid)
- 
---Watch acpi state when booting a VM, used to be handled in xenvm
-monitorAcpi :: Uuid -> VmMonitor -> AcpiState -> IO ()
-monitorAcpi uuid m state = do
-    acpi_state <- Xl.acpiState uuid
-    if state == 5
-      then do return ()
-      else do
-        if state /= acpi_state
-          then do vmStateSubmit m acpi_state
-                  threadDelay (10^6)
-                  monitorAcpi uuid m acpi_state
-          else do threadDelay (10^6)
-                  monitorAcpi uuid m acpi_state
 
 -- Creates a snapshot from the primary disk when the disk persistence option is
 -- set. It first checks to see if a snapshot already exists and deletes it before
@@ -887,7 +873,6 @@ bootVm config reboot
        -- ensure bootstrap and phase handling synchronously terminates before returning (and errors get propagated)
        force bootstrap
        force phases
-       liftIO . void $ forkIO $ monitorAcpi uuid monitor 0
 
       writable domid path = do
         xsWrite path ""
--- a/xenmgr/Vm/Actions.hs
+++ b/xenmgr/Vm/Actions.hs
@@ -775,6 +775,10 @@ setupAcpiNode uuid =
      case stubdom of
          Just stubdomid -> liftIO $ xsChmod (xsp domid ++ "/acpi-state") ("b" ++ show stubdomid)
          Nothing        -> liftIO $ xsChmod (xsp domid ++ "/acpi-state") ("b" ++ show domid)
+     liftIO $ xsWrite (xsp domid ++ "/power-state") ("")
+     case stubdom of
+         Just stubdomid -> liftIO $ xsChmod (xsp domid ++ "/power-state") ("b" ++ show stubdomid)
+         Nothing        -> liftIO $ xsChmod (xsp domid ++ "/power-state") ("b" ++ show domid)
 
 -- Creates a snapshot from the primary disk when the disk persistence option is
 -- set. It first checks to see if a snapshot already exists and deletes it before
--- a/xenmgr/Vm/Monitor.hs
+++ b/xenmgr/Vm/Monitor.hs
@@ -71,6 +71,7 @@ data VmEvent
    | VmMeasurementFailure !FilePath !Integer !Integer
    | VmStateUpdate
    | VmAcpiUpdate
+   | VmPowerUpdate
      deriving (Eq,Show)
 
 data VmMonitor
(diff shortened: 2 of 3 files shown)


jandryuk:
> Removing the thread noticably reduces xenmgr's CPU usage in top.

xenmgr at like 1.x% CPU usage, but with this change it drops to ~0%.  The bigger decrease is the whole vm in xentop.  xentop shows dom0 drops from 6-9% to ~1.5% when otherwise idle.  The forked xl processes don't show up in top, but you see them in xentop's aggregate CPU usage.

> xenmgr can basically idle for 30 seconds between polling the disk usage.

To be clear, by removing monitorAcpi, xenmgr can idle for ~30 seconds at a time.

xenmgr/Vm/Actions.hs
```diff
@@ -775,20 +775,6 @@ setupAcpiNode uuid =
      case stubdom of
          Just stubdomid -> liftIO $ xsChmod (xsp domid ++ "/acpi-state") ("b" ++ show stubdomid)
          Nothing        -> liftIO $ xsChmod (xsp domid ++ "/acpi-state") ("b" ++ show domid)
- 
---Watch acpi state when booting a VM, used to be handled in xenvm
```
crogers1, line 779: Removing this would also drop the ```powerlinkR``` stuff, which is supposed to change the host state based on guest state.  So if you sleep the windows VM, it was supposed to sleep the host.  Whether or not that's working right now is unknown, and whether we want to continue to 'support' it is also up for debate.

You're right for the most part about the rest of the state change stuff, the important callbacks are fired based off vm state, not acpiState, like ```set_shr``` which is needed at the moment to differentiate between reboot and shutdown.  However one thing it is useful for is performing an action when the guest sleeps or resumes from sleep, since the VM state is always considered 'running'. Only the acpiState denotes that it is asleep/awake.

jandryuk:
https://github.com/OpenXT/manager/pull/192/commits/e7811ac57919d4560052aae20fab11910f6e260c re-adds S4/Hibernated tracking.

Since this has been broken for a while, I'm okay with deleting code instead of adding this back in.

xenmgr/Vm/Actions.hs
crogers1: We do have guest S3 working, we'll try and make sure whatever is making that work gets upstreamed soon.  

This PR got away from me, I'll need to look through it again but I don't disagree with what you've said here.  I'm definitely not married to the powerlinkR stuff, although it would be nice to evaluate whether we have any more luck with host S3 on a newer xen version.  I've completely forgotten the reason why it doesn't work today.

jandryuk: You wrote:

> We do have guest S3 working, we'll try and make sure whatever is making that work gets upstreamed soon.

When you say guest, you mean VM?  I note these two failures in the PR description:

> Also, VM S3 seems broken. Windows 10 can't S3 as an OpenXT VM because the graphics subsystem prevents it. However, I haven't checked if the vglass drivers allow it. Linux detaches its PV disks on S3 resume, but the backend doesn't reconnect so it is basically dead.

Windows S3 disabling (shown by `powercfg /a`) is with emulated VGA.  Maybe it works for you with the vGlass PV drivers?

jandryuk:
I just tested setting a VM "Power Control" to "Power control affects VM and host".  Shutting down that VM shutdown the host, so I think that shows powerlink is working.

With these three commits together:
[xenmgr: Handle s0 in reactVmAcpiUpdate](https://github.com/OpenXT/manager/pull/192/commits/6eadd3572a27a3f06fd8cbaa3f5878592c9bb522)
[xenmgr: Delete monitorAcpi](https://github.com/OpenXT/manager/pull/192/commits/a1425c815592e91956b0cc586414d6ee76070fe4)
[xenmgr: Sythensize VM ACPI state changes](https://github.com/OpenXT/manager/pull/192/commits/04c0e0d9b18ee06711bbd5d28026d8ca4064842b)
I don't think anything has regressed.

This restores setting a VM as hibernated, so the UI says it is hibernated:
[xenmgr: Wire-up power-state tracking](https://github.com/OpenXT/manager/pull/192/commits/e7811ac57919d4560052aae20fab11910f6e260c)
But since it's been broken for a while, I wonder if we would be better off dropping it and then removing the qemu-dm writing of power-state.  Also our installation instructions recommend disabling hibernate.

jandryuk:
With: [xenmgr: Wire-up power-state tracking](https://github.com/OpenXT/manager/pull/192/commits/e7811ac57919d4560052aae20fab11910f6e260c) + powerlink, I hibernated a Win10 guest - it and the host shutdown.  Powering back on, the Win10 guest was restarted.  xenmgr wrote "coming out of hibernation, vms to start: [...$uuids ]".  uuids was uivm, 3 ndvms, and the win10 guest.

I think there is a slight wrinkle.  Win10 by default hibernates and doesn't shutdown.  The restored session didn't have a logged in user.  If you disabled hibernate then shutdown shuts down.  Explicitly hibernating the guest works as expected.  It's that with powerlink, with the default hibernate enabled setting, clicking shutdown from outside is a hibernate.  So when you reboot, that VM is restored.  Start on boot was not set, so it was getting restored via the faux-shutdown->hibernate.  OpenXT is doing the right thing - it's the Windows behavior which is surprising.

crogers1 (approved):
I think I'm good with these changes.  Dropping the monitor thread is a plus, and re-wiring the power-state logic is also good.

The windows behavior for power state ops should be set and confirmed by the user/admin for any powerlink configurations.

Merging soon.

PR #195, opened 2022-09-13
xenmgr: Fix enable-dom0-networking

jandryuk:
Trying to disable dom0 networking by setting xenmgr.config propery enable-dom0-networking to False fails with:
223:IO error: /config/system/dom0-networking-disabled: openFile: does not exist (No such file or directory)

/config/system does not exist.  Ensure we create it if needed.

### DIFF
--- a/xenmgr/XenMgr/Config.hs
+++ b/xenmgr/XenMgr/Config.hs
@@ -72,6 +72,7 @@ import qualified Control.Exception as E
 import Data.List
 import Data.Maybe
 import Data.Char
+import System.Directory
 import System.IO
 import System.IO.Unsafe
 import System.Posix.Files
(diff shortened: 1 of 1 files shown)


xenmgr/XenMgr/Config.hs
```diff
@@ -269,7 +270,9 @@ appSetEnableDom0Networking True =
       exist <- doesFileExist "/config/system/dom0-networking-disabled"
       if exist then removeFile "/config/system/dom0-networking-disabled" else return ()
 appSetEnableDom0Networking False =
-    liftIO $ writeFile "/config/system/dom0-networking-disabled" ""
+    liftIO $ do
+      createDirectoryIfMissing True "/config/system"
```
crogers1, line 274: This must have relied on ```/config/system``` being created somewhere else.  Tough to say where exactly but this is 100% fine to add as a defensive check.

crogers1 (approved):
LGTM.

PR #198, opened 2022-11-07
xenmgr: Re-try mount multiple times

jandryuk:
We see consistent failure of a PV NDVM to boot where xenmgr fails to mount the VHD partition.  /dev/loop0p1 is still labeled device_t, so mount cannot mount it.

This adds a re-try loop to try again instead of failing.

The addition on udevadm settle was an attempt to address this, but it is proving insufficient for some reason.

@crogers1 - Is there no ordering in:
```
                   settle part
                   mount loop temp_dir ro
```

i.e. the intent was to have settle complete before mount is called, but they may be run in any order?

### DIFF
--- a/xenmgr/Vm/Utility.hs
+++ b/xenmgr/Vm/Utility.hs
@@ -60,6 +60,18 @@ mount dev dir ro = void $ readProcessOrDie "mount" ["-o", opts, dev, dir] "" whe
   ro_opt | ro = "ro"
          | otherwise = ""
 
+mount_try_n n dev dir ro =
+  do
+    info $ "mount " ++ (show n) ++ " " ++ dev ++ " on " ++ (show dir) ++ " read-only:" ++ (show ro)
+    readProcessOrDie "udevadm" ["settle"] ""
+    mount dev dir ro `E.catch` mount_error n
+  where
+    mount_error :: Int -> E.SomeException -> IO ()
+    mount_error 0 e = E.throw e
+    mount_error i e = do mount_try_n (i - 1) dev dir ro
+
+mount_try = mount_try_n 7
+
 umount :: FilePath -> IO ()
 umount dir = void $ readProcessOrDie "umount" [dir] ""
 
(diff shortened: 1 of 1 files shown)


xenmgr/Vm/Utility.hs
```diff
@@ -60,6 +60,18 @@ mount dev dir ro = void $ readProcessOrDie "mount" ["-o", opts, dev, dir] "" whe
   ro_opt | ro = "ro"
          | otherwise = ""
 
+mount_try_n n dev dir ro =
+  do
+    info $ "mount " ++ (show n) ++ " " ++ dev ++ " on " ++ (show dir) ++ " read-only:" ++ (show ro)
```
crogers1, line 65: I would add a 
```haskell
threadDelay (10^6)
```
just to give it some time.  I bet there's some inherent delay waiting for udevadm settle to return, but there's nothing wrong with pausing explicitly for a second on each retry loop.

crogers1:
Both those functions call ```createProcess``` under the hood and wait for the underlying pids to exit with a call to ```waitForProcess`` in xclibs/xchutils/Tools/Process.hs:179, so they should be synchronous.  If that's not the behavior you're seeing maybe there's some nuance with the subshell we're not taking into account, but I can't think of anything off the top of my head.

Either way, a retry loop is totally fine.

jandryuk:
@crogers1 

So while system/createProcess is synchronous, is Haskell free to re-order:
```
                   settle part
                   mount loop temp_dir ro
```
to
```
                   mount loop temp_dir ro
                   settle part
```
?

Do you need something like `first >> second` to ensure ordering?

I'd rather not add a delay...  this really shouldn't take a second, though I suppose I could use a smaller delay.

crogers1:
Those two lines are already in `do` notation, which implies `first >> second`, it shouldn't be doing any reordering.  Were you seeing the mount happen before the udevadm settle?

crogers1:
If it's OBE let's close it.

jandryuk:
Closing.  It's better to not use this code path, and we don't by default when booting via grub-xen-pvh/pv64.

PR #197, opened 2022-10-13
[xenmgr] Add idle shutdown specific call

crogers1:
Implement a separate function for handling the idle shutdown
  case. When the system has been idle for the specified idle period
  and no user input has been detected, shutdown the platform. Start
  a 5 minute timer that will force the host off if has not completed
  the shutdown process.

  If the idle timer feature is enabled, then the expected behavior is
  that the host will be shutdown. This commit ensures that the host will
  reach the off state in the event of a stuck guest VM or otherwise non-
  responding process when handling an idle shutdown within a reasonable
  amount of time.

Signed-off-by: Chris Rogers <rogersc@ainfosec.com>

PR #199, opened 2022-11-07
Remove xenmgr idle support

jandryuk:
xenmgr has a thread that polls vglass for idle time every 30 seconds.  vglass hardcodes the response to 0, so it is ineffective.  xcpmd has idle time support after https://github.com/OpenXT/manager/pull/197 and https://github.com/OpenXT/xctools/pull/74, so this loop can be removed.

The second commit makes the xenmgr config.ui properties hardcoded.  The UI still has some code to query them.  That could be removed with futher idl and toolstack-data clean ups.  I didn't want to deal with the rebuild from changing the idl, so I didn't attempt that.

At least the first patch is nice since it removes pointless polling and eliminates log spam in console mode (runlevel 3).


crogers1 (approved):
This LGTM.  Merging later.

PR #200, opened 2022-11-07
xenmgr: Push ACPI power button multiple times

jandryuk:
Sometimes the shutdown command does not work.  It seems like the longer the VM is up, the more likely it is to not shutdown.  This is based on Windows 10 without PV drivers.

The button press is seen by the VM to some extent because a DPMS off (black) screen will turn on, but shutdown is not initiated.  A second press will trigger it - that was the intent to the xl trigger power line added in commit 1cfb6aa8e63e "xenmgr: Add xl trigger power to HVM shutdown".

Re-work the code so a background thread is started that will push the power buttons* 3 times each with a 1 second delay.  This will hopefully let the VM recognize the button press without going on for too long.

*HVMs have two power and two sleep buttons.  One is Xen emulating buttons for HVMs, and the second is from QEMU's acpi-pm-features.patch and ACPI changes.  xl trigger power is pushing the Xen one and the xenstore hvm-shutdown write is triggering the QEMU one.

With forkIO, the `xl shutdown -F -ww` runs before the `xl trigger power` commands.  xl shutdown first tries PV shutdown and then (-F) fallback to ACPI.  So this would push the Xen ACPI button before the QEMU one in pushPowerButton.

This was previously included with https://github.com/OpenXT/manager/pull/193 but can go in independently.  I wrote the above a while ago.  I think the scenario is Windows without PV drivers.  The screen has blanked via DPMS.  A single power button press takes it out of DPMS.  A second power button press then initiates shutdown.  (The VM needs to be configured to shutdown on power button press).  So 3 power button presses is more than seems to be strictly necessary, but it's what I wrote and have tested.  It doesn't seem to hurt anything to do it more times.

### DIFF
--- a/xenmgr/XenMgr/Connect/Xl.hs
+++ b/xenmgr/XenMgr/Connect/Xl.hs
@@ -166,6 +166,22 @@ domainXsPath uuid = do
       "" -> return $ "/local/domain/unknown"
       _  -> return $ "/local/domain/" ++ domid
 
+pushPowerButton :: Uuid -> Int -> IO ()
+pushPowerButton uuid count = do
+      domid <- getDomainId uuid
+      stubdomid <- getStubDomainID uuid
+      let xs_path = "/local/domain/" ++ stubdomid ++ "/device-model/" ++ domid
+      _pushPowerButton uuid domid xs_path 1 count
+    where
+      _pushPowerButton :: Uuid -> String -> String -> Int -> Int -> IO ()
+      _pushPowerButton uuid domid xs_path i max = do
+          debug $ "push power button " ++ show uuid ++ " " ++ show i ++ " of " ++ show max
+          xsWrite (xs_path ++ "/hvm-shutdown") "poweroff"
+          system_ ("xl trigger " ++ domid ++ " power")
+          if i < max
+              then do threadDelay $ 10^6
+                      _pushPowerButton uuid domid xs_path ( i + 1 ) max
+              else return ()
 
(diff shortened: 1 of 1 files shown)


crogers1:
Still looking at this, but we might be better off just doing a switch and possibly injecting an input event like a mouse move or ctrl key press instead of sending shutdown multiple times.  If I can propose an easy way to do this I will.

jandryuk:
FYI, redfield would push the button once a second until the vm shutdown.

crogers1 (approved):
Alrighty.  I can't recall (and don't have a box in front of me atm), but does ```xec-vm shutdown``` block until the domain is off?  If not then then the forkIO is fine.

jandryuk:
`xec-vm shutdown` (the equivalent dbus call in xenmgr) blocks until the domain powers off.

The `-ww` in `system_ ("xl shutdown -ww " ++ domid)` has xl wait.  That call fails if PV drivers are not available, in which case `_ <- system_ ("xl shutdown -F -ww " ++ domid)` does the -F fallback ACPI shutdown and waits.

The `xl trigger power` calls do not wait, and the xsWrites don't wait either.  Those get re-worked into the pushPowerButton.  I guess if the domain shuts down faster than 2 seconds, there could be calls for a stale domain.  However, this path is only for non-PV driver domains, so Windows without PV drivers.  AFAIK, Windows can't shutdown that fast :)  You'd need the domid to be re-used in a small window, and it would be just to issue a shutdown, so I think it's okay.

PR #201, opened 2023-03-13
Hash VHD files

jandryuk:
Today, xenmgr hashes VHDs by creating tapdevs and hashing across the exposed contents.  This was done because `tap-ctl create` modifies .vhd metadata by default and hashing the .vhd file itself would be inconsistent.

If read-only tapdevs are created with `tap-ctl create -R`, the .vhd will not be modified.  By making xenmgr use read-only tapdevs, we can change to hashing across the .vhds.

This does mean that any use of `tap-ctl create` without `-R` would mess up the hashes.  A separate change to make `tap-ctl create` default to read-only, with `-W` to create writable, is forthcoming.


crogers1 (approved):
Interested in seeing the tap-ctl changes but from a code perspective this looks fine.

jandryuk:
This is the corresponding change for read-only by default: https://github.com/OpenXT/xenclient-oe/pull/1466

crogers1 (approved):
LGTM. Build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/80

PR #202, opened 2023-03-22
xenmgr: Fix initrd support

jandryuk:
initrd/initramfs support is broken today.  xenmgr writes the xl.cfg with "initrd=", but the proper xl.cfg entry is "ramdisk=".  Fix that.

If an HVM sees ramdisk= without a kernel specified, you get this failure:
xl: [5605] libxl_create.c:169:libxl__domain_build_info_setdefault:direct boot parameters specified but kernel missing

So move the ramdisk= setting into virtSpecs and only run it for PVH and PV.


crogers1 (approved):
LGTM. Build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/80

PR #204, opened 2023-03-22
xenmgr: Let sound=none disable it

jandryuk:
Let sound=none disable audio device for a VM.  Otherwise the dbd entry needs to be manually deleted to remove the audio device.


crogers1 (approved):
LGTM. Build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/79

PR #205, opened 2023-03-22
templates: Use grub-xen-pvh/pv64 for VMs

jandryuk:
Stop extracting the kernel and let these VMs boot via grub-xen-pvh or grub-xen-pv64 respectively.

### DIFF
--- a/templates/default/service-ndvm-pv
+++ b/templates/default/service-ndvm-pv
@@ -47,7 +47,7 @@
     "memory": "192",
     "display": "none",
     "cmdline": "root=\/dev\/xvda2 iommu=soft console=hvc0",
-    "kernel-extract": "0,1:\/bzImage",
+    "kernel": "\/usr\/lib\/xen\/boot\/grub-xen-pv64",
     "flask-label": "system_u:system_r:ndvm_t",
     "pci": {
       "0": {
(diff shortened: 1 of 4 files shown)


crogers1:
Build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/80

templates/default/service-ndvm-pv
```diff
@@ -47,7 +47,7 @@
     "memory": "192",
     "display": "none",
     "cmdline": "root=\/dev\/xvda2 iommu=soft console=hvc0",
-    "kernel-extract": "0,1:\/bzImage",
+    "kernel": "\/usr\/lib\/xen\/boot\/grub-xen-pv64",
```
crogers1, line 50: You're going to need migration rules here to clear `kernel-extract`, and set `kernel`, for existing VMs on upgrade.  `upgrade-db/Migrations/M_37.hs` is a good reference.

jandryuk: I had kinda thought I could get away without one.  The new service-ndvm and service-uivm would handle.  Existing ndvms and syncvms are already working and should keep working, so they don't need to change.

If the disk image were copied, then it wouldn't be updated with an appropriate grub.cfg.  So I guess a heuristic of "if type==ndvm and disk path==/storage/ndvm/ndvm.vhd, then update kernel config" would work.

Thanks for the review.

crogers1: Welcome.  If there are service VMs configs that don't utilize the "service-" template feature (that overwrites the db entries every boot), then they need a migration rule.

crogers1 (approved):

crogers1:
LGTM.  Merging soon.

PR #206, opened 2023-04-04
templates: sync ndvm-pv and ndvm-hvm

jandryuk:
Since commit 5d1bcbac43ee "xenmgr: Avoid generating invalid configs" a unified config can be used and switched by just flipping virt-type. Bring the two configs together so they just differ by virt-type.


crogers1 (approved):

crogers1:
Merging soon.

PR #208, opened 2023-10-31
xenmgr: Handle missing soundcards

jandryuk:
When there is no sound card, the UI shows IO Error, and log files show: IO error: /proc/asound/pcm: openFile: does not exist (No such file or directory)

Catch the open errors (inside readFile) and return an empty string. Subsequent parsing won't match and there will be no devices returned.

Also handle /proc/asound/cards missing, which could occur if dom0 isn't built with sound card support.


crogers1 (approved):
LGTM. Merging soon.

